PR #67, opened 2019-07-19
Compute derivatives of the tree likelihood wrt branch lengths.

4ment:
Compute derivatives of the tree likelihood wrt branch lengths.
Implemented a test on a single tree.

### DIFF
--- a/src/libsbn.hpp
+++ b/src/libsbn.hpp
@@ -138,6 +138,31 @@ struct SBNInstance {
     return results;
   }
 
+  std::vector<std::vector<double>> BranchGradients() {
+    if (beagle_instances_.size() == 0) {
+      std::cerr << "Please add some BEAGLE instances that can be used for "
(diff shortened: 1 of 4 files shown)


src/libsbn.hpp
```diff
@@ -138,6 +138,31 @@ struct SBNInstance {
...
+    }
+    std::vector<std::vector<double>> results(tree_collection_->TreeCount());
+    std::queue<beagle::BeagleInstance> instance_queue;
+    for (auto instance : beagle_instances_) {
+      instance_queue.push(instance);
+    }
+    std::queue<size_t> tree_number_queue;
+    for (size_t i = 0; i < tree_collection_->TreeCount(); i++) {
+      tree_number_queue.push(i);
+    }
```
matsen: I guess we can put the above into some separate function to avoid repetition here.

PR #107, opened 2019-09-30
106 gradient hmc clock

christiaanjs:
Includes implementation of branch length gradients and custom conda recipe for BEAGLE branch

### DIFF
--- a/beagle-recipe/meta.yaml
+++ b/beagle-recipe/meta.yaml
@@ -1,0 +1,35 @@
+{% set name = "beagle-lib" %}
+{% set version = "3.1.2" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  git_url: git@github.com:beagle-dev/beagle-lib.git
+  git_tag: hmc-clock
+
+build:
+  number: 1
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - autoconf
+    - automake
+    - libtool
+    - openjdk
+  run:
+    - openjdk
+    - libtool
+
+test:
+  commands:
+    - test -f ${PREFIX}/lib/libhmsbeagle-jni.so # [linux]
+    - test -f ${PREFIX}/lib/libhmsbeagle-jni.jnilib # [osx]
+
+about:
+  home: https://github.com/beagle-dev/beagle-lib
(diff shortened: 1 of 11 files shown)


beagle-recipe/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+build:
+  number: 1
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - autoconf
+    - automake
+    - libtool
+    - openjdk
```
matsen: It always seemed strange to me that the beagle conda required openjdk. Have you tried without?

src/beagle.cpp
```diff
@@ -388,4 +408,125 @@ std::vector<std::pair<double, std::vector<double>>> BranchGradients(
...
+  for (int index : derivative_matrix_indices) {
+    beagleSetDifferentialMatrix(beagle_instance, index, Q.data());
+  }
+
+  beagleUpdatePartials(beagle_instance, operations.data(),
+                       static_cast<int>(operations.size()),
+                       BEAGLE_OP_NONE);  // cumulative scale index
+
+  // Calculate pre-order partials
+  // TODO: Set root pre-order partials equal to root state frequencies
```
matsen: Is this still outstanding?

beagle-recipe/meta.yaml
christiaanjs: It looks like a JDK is an optional dependency for Beagle so I will remove it

christiaanjs:
The new branch length gradient implementation is forked from @4ment . Should I go ahead and replace the existing implementation with it/tidy it up or leave it to him? Also, it doesn't look like I have push permissions to the `phylovi/libsbn` repository anymore.

christiaanjs:
Also, I've added the setup process to the Dockerfile but I'm not very familiar with it so I'm not sure how to get the Travis build to pass

matsen:
I just chatted with @4ment who suggests that you replace the existing implementation.

Regarding the Travis process, you should be able to simulate it by running 

    docker build -t phylovi/libsbn .

then 

    docker run -v ${PWD}:/libsbn -t phylovi/libsbn /bin/bash -c "/opt/conda/bin/conda run -n libsbn make -C /libsbn"

PR #184, opened 2020-02-26
181 scale em

junseonghwan:
- Removed if statement that checks for LOG_EPS for probability of a rooted topology, which was the cause that led to the EM algorithm failing to update the parameters for large trees (all rooted trees had probability lower than LOG_EPS so nothing was being updated).
- Introduced ERR_TOLERANCE = 1e-10. This value is used rather than EPS to check for potential reversion of EM score in the subsequent iterations . EPS represents the smallest representable value close to 0 and it turned out to be too strict, resulting in failed assertions when the difference in score from current iteration to the previous iteration was something like -1e-16.
- Overflow/underflow problem stemming from summing the log probabilities is handled by checking for floating point environment and assigning DOUBLE_MINIMUM when it occurs, which is defined as the lowest representable double times ERR_TOLERANCE=1e-10. This tolerance value is needed as the code may continue to perform operations on DOUBLE_MINIMUM, for example via LogSum/LogAdd. When using lowest double, it resulted in further floating point errors leading to numerical instability.

### DIFF
--- a/src/sbn_probability.cpp
+++ b/src/sbn_probability.cpp
@@ -95,11 +95,15 @@ double ProductOf(const EigenConstVectorXdRef vec, const SizeVector& indices,
 }
 
 // Take the sum of the entries of vec in indices plus starting_value.
-double SumOf(const EigenConstVectorXdRef vec, const SizeVector& indices,
+// The contents of vec is log of probabilities, i.e., they are negative values.
+double SumOfLogProbabilities(const EigenConstVectorXdRef vec, const SizeVector& indices,
              const double starting_value) {
   double result = starting_value;
   for (const auto& idx : indices) {
     result += vec[idx];
+    if (result < DOUBLE_MINIMUM) {
+      return DOUBLE_MINIMUM;
+    }
   }
   return result;
 }
@@ -233,25 +237,24 @@ EigenVectorXd SBNProbability::ExpectationMaximization(
       // Loop over the various rooting positions of this topology, using log_q_weights
       // to store the probability of the tree in the various rootings (we will normalize
       // it later).
+      double log_q_weight_sum = DOUBLE_NEG_INF;
+      NumericalUtils::ReportFloatingPointEnvironmentExceptions("|Before Rooting|");
       for (size_t rooting_position = 0; rooting_position < edge_count;
            ++rooting_position) {
         const SizeVector& rooted_representation =
             indexer_representation[rooting_position];
         // Calculate the SBN probability of this topology rooted at this position.
-        double log_p_rooted_topology = SumOf(sbn_parameters, rooted_representation, 0.);
-        log_q_weights[rooting_position] =
-            (log_p_rooted_topology > LOG_EPS) ? log_p_rooted_topology : LOG_EPS;
+        double log_p_rooted_topology = SumOfLogProbabilities(sbn_parameters, rooted_representation, 0.);
+        log_q_weights[rooting_position] = log_p_rooted_topology;
       }  // End of looping over rooting positions.
-      // This is the likelihood of the topology, marginalized over rooting position.
-      double log_p_unrooted_topology = NumericalUtils::LogSum(log_q_weights);
-      // Increase the total log marginal likelihood by this topologies contribution.
-      score_history[em_idx] += topology_count * log_p_unrooted_topology;
+      NumericalUtils::ReportFloatingPointEnvironmentExceptions("|After Rooting|");
+      log_q_weight_sum = NumericalUtils::LogSum(log_q_weights);
+      score_history[em_idx] += topology_count * log_q_weight_sum;
       // Normalize q_weights to achieve the E-step of Algorithm 1.
       // For the increment step (M-step of Algorithm 1) we want a full topology
       // count rather than just the unique count. So we multiply the q_weights by the
       // topology count (in log space, it becomes summation rather than multiplication).
-      log_q_weights =
-          log_q_weights.array() + (-log_p_unrooted_topology + log(topology_count));
+      log_q_weights = log_q_weights.array() + (-log_q_weight_sum + log(topology_count));
       // Increment the SBN-parameters-to-be by the q-weighted counts.
       IncrementByInLog(log_m_bar, indexer_representation, log_q_weights);
     }  // End of looping over topologies.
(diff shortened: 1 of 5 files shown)


src/sbn_probability.cpp
```diff
@@ -232,25 +236,32 @@ EigenVectorXd SBNProbability::ExpectationMaximization(
...
         const SizeVector& rooted_representation =
             indexer_representation[rooting_position];
         // Calculate the SBN probability of this topology rooted at this position.
-        double log_p_rooted_topology = SumOf(sbn_parameters, rooted_representation, 0.);
-        log_q_weights[rooting_position] =
-            (log_p_rooted_topology > LOG_EPS) ? log_p_rooted_topology : LOG_EPS;
+        double log_p_rooted_topology = SumOfLogProbabilities(sbn_parameters,
+                                                     rooted_representation, 0.);
+        // SHJ: Sometimes overflow is reported, sometimes it's underflow...
+        if (fetestexcept(FE_OVERFLOW) || fetestexcept(FE_UNDERFLOW)) {
```
matsen: You can combine this and the feclearexcepts into a single statements by defining a variable for `FE_OVERFLOW | FE_UNDERFLOW`. e.g. see definition of `FE_WORRYING_EXCEPT`.

src/sbn_probability.cpp
```diff
@@ -232,25 +236,32 @@ EigenVectorXd SBNProbability::ExpectationMaximization(
...
+        } else {
+          log_q_weights[rooting_position] = log_p_rooted_topology;
+        }
       }  // End of looping over rooting positions.
-      // This is the likelihood of the topology, marginalized over rooting position.
-      double log_p_unrooted_topology = NumericalUtils::LogSum(log_q_weights);
-      // Increase the total log marginal likelihood by this topologies contribution.
-      score_history[em_idx] += topology_count * log_p_unrooted_topology;
+      NumericalUtils::ReportFloatingPointEnvironmentExceptions("|After Rooting|");
+      log_q_weight_sum = NumericalUtils::LogSum(log_q_weights);
```
matsen: I feel like the `log_p_unrooted_topology` that was in the previous version is actually a more descriptive name. What do you think?

src/sbn_probability.cpp
```diff
@@ -264,24 +275,37 @@ EigenVectorXd SBNProbability::ExpectationMaximization(
       // Last line of the section on EM in doc/tex.
       score_history[em_idx] += m_tilde_for_positive_alpha.dot(sbn_parameters);
     }
+    // SHJ: Print the score in the below line.
+    // Uncommenting it but leaving it incase we need to do further debugging.
+    // Remove it at some point.
+    //std::cout << std::setprecision(20) << score_history[em_idx] << std::endl;
```
matsen: I like to take out code comments like this but put them as comments on the GH pull request.

src/sbn_probability.cpp
```diff
@@ -264,24 +275,37 @@ EigenVectorXd SBNProbability::ExpectationMaximization(
...
       double scaled_score_improvement =
           (score_history[em_idx] - score_history[em_idx - 1]) /
           fabs(score_history[em_idx - 1]);
       // To monitor correctness of EM, we check to ensure that the score is
       // monotonically increasing (modulo numerical instability).
-      Assert(scaled_score_improvement > -EPS, "Score function decreased.");
+      // SHJ: -EPS is too big, I noticed the assertion failure for
+      // scaled_score_improvement of -6e-16. Using ERR_TOLERANCE
+      // which is 1e-10.
+      //Assert(scaled_score_improvement > -EPS, "Score function decreased.");
```
matsen: That's fine, but let's take out the commented code.

src/sbn_probability.cpp
```diff
@@ -264,24 +275,37 @@ EigenVectorXd SBNProbability::ExpectationMaximization(
...
     if (em_idx > 0) {
       double scaled_score_improvement =
           (score_history[em_idx] - score_history[em_idx - 1]) /
           fabs(score_history[em_idx - 1]);
       // To monitor correctness of EM, we check to ensure that the score is
       // monotonically increasing (modulo numerical instability).
-      Assert(scaled_score_improvement > -EPS, "Score function decreased.");
+      // SHJ: -EPS is too big, I noticed the assertion failure for
+      // scaled_score_improvement of -6e-16. Using ERR_TOLERANCE
+      // which is 1e-10.
```
matsen: I'd prefer not to have a comment describing the value of a define, because the value might change and the comment then becomes out of date.

matsen:
Here's a block of code that was once useful:

```cpp
  auto finite_count = std::count_if(sbn_parameters.begin(), sbn_parameters.end(),
                                    [](double d) { return std::isfinite(d); });
  std::cout << finite_count << " / " << sbn_parameters.size()
```

src/sbn_probability.cpp
```diff
@@ -94,7 +97,8 @@ double ProductOf(const EigenConstVectorXdRef vec, const SizeVector& indices,
 }
 
 // Take the sum of the entries of vec in indices plus starting_value.
-double SumOf(const EigenConstVectorXdRef vec, const SizeVector& indices,
+// The contents of vec is log of probabilities, i.e., they are negative values.
+double SumOfLogProbabilities(const EigenConstVectorXdRef vec, const SizeVector& indices,
```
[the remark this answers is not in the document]
junseonghwan: Nope. This function can go back to just being SumOf.

src/sbn_probability.cpp
```diff
@@ -232,25 +236,32 @@ EigenVectorXd SBNProbability::ExpectationMaximization(
       // Loop over the various rooting positions of this topology, using log_q_weights
       // to store the probability of the tree in the various rootings (we will normalize
       // it later).
+      double log_q_weight_sum = DOUBLE_NEG_INF;
+      NumericalUtils::ReportFloatingPointEnvironmentExceptions("|Before Rooting|");
```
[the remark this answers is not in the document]
junseonghwan: I'll remove them.

src/sbn_probability.cpp
junseonghwan: Done.

src/sbn_probability.cpp
junseonghwan: Done: log_q_weight_sum changed back to log_p_unrooted_topology.

src/sbn_probability.cpp
junseonghwan: Taken out.

src/sbn_probability.cpp
junseonghwan: Removing commented Assert and 1e-10 from the comments.

src/sbn_probability.cpp
junseonghwan: Removed 1e-10.

Issue #187, opened 2020-02-29
Node height gradients for RootedTrees


matsen:
* [x] Add appropriate infrastructure to RootedTree to enable gradients
* [x] Consider renaming `parameters_`, which seems too vague
* [x] Gradients!
* [x] Unit tests in RootedSBNInstance
* [x] Perhaps unit tests in RootedTree?
* [ ] Drop NoClockModel (if not gone already)

matsen:
Here is an image that goes with the RootedTree unit test:

![IMG_20200423_070609](https://user-images.githubusercontent.com/112708/80109526-71a94880-8532-11ea-8f1a-020fb6247a9d.jpg)

matsen:
Closed by https://github.com/phylovi/libsbn/commit/fb89cbd71952851681ce23670c6d9b44cf1291a1

PR #203, opened 2020-04-14
Node height gradients

matsen:
See issue #187


src/libsbn.hpp
```diff
@@ -392,6 +392,34 @@ TEST_CASE("libsbn: likelihood and gradient") {
   }
 }
 
+#include "rooted_sbn_instance.hpp"
```
matsen, line 395: This should be moved to `rooted_sbn_instance.hpp` -- there is a "dummy" test there that you can replace.

README.md
```diff
@@ -127,9 +127,9 @@ Add a test for every new feature.
 * Seong-Hwan Jun ([@junseonghwan](https://github.com/junseonghwan)): implementation of SBN gradients, design
 * Cheng Zhang ([@zcrabbit](https://github.com/zcrabbit)): concept, design, algorithms
 * Christiaan Swanepoel ([@christiaanjs](https://github.com/christiaanjs)): design
+* Xiang Ji ([@xji3](https://github.com/xji3)): gradient expertise and node height code
+* Marc Suchard ([@msuchard](https://github.com/msuchard)): gradient expertise and node height code
```
matsen, line 131: Do you want to put something in the Acknowledgements below about how parts of the gradient code were adapted from BEAGLE?

src/rooted_tree.hpp
```diff
@@ -15,7 +15,8 @@ class RootedTree : public Tree {
   bool operator==(const Tree& other) const = delete;
   bool operator==(const RootedTree& other) const;
 
- private:
+  void InitializeParameters(const std::unordered_map<Tag, double>& taxon_date_map);
```
matsen, line 18: Rename to `tag_date_map`, please.

src/rooted_tree.cpp
```diff
@@ -9,6 +9,47 @@ RootedTree::RootedTree(const Tree& tree) : Tree(tree.Topology(), tree.BranchLeng
...
+
+  node_heights_ = std::vector<double>(Topology()->Id() + 1);
+  node_bounds_ = std::vector<double>(Topology()->Id() + 1);
+
+  std::vector<double> branch_lengths = BranchLengths();
+  int root_id = static_cast<int>(Topology()->Id());
+
+  // Set up bounds and heights
+  // Start with the leaves from the dates map
+  for (const auto& pair : taxon_date_map) {
```
matsen, line 25: Remember that in C++17 we can introduce meaningful names and avoid first and second with things like
```
for (const auto &[begin, end] : subsplit_ranges) {
```

src/fat_beagle.cpp
```diff
@@ -73,10 +73,10 @@ double FatBeagle::LogLikelihood(const RootedTree &tree) const {
 }
 
 std::pair<double, std::vector<double>> FatBeagle::BranchGradientInternals(
-    const Tree &tree) const {
+    const Tree &tree, const std::vector<double> &branch_lengths) const {
```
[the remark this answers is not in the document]
4ment: In a rooted tree branch lengths are in years while in the unrooted case they represent the expected number of substitutions. Trees are const in this class so branch_lengths_ cannot be modified to take the substitution rate into account.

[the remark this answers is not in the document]
matsen: Is it clear that I'm proposing changing the signature to use a `const Node::NodePtr` (i.e. a topology) rather than a `const Tree &`?

README.md
4ment: It could be added to the code as a comment

src/beagle_accessories.hpp
```diff
@@ -46,14 +46,14 @@ struct BeagleAccessories {
   // first derivative matrices
   std::vector<int> node_deriv_index_ = {0};
 
-  BeagleAccessories(int beagle_instance, bool rescaling, const Tree &tree)
+  BeagleAccessories(int beagle_instance, bool rescaling, const Node::NodePtr root)
```
matsen, line 49: I like this type change. 

Elsewhere in the code (e.g. https://github.com/phylovi/libsbn/blob/master/src/libsbn.cpp#L127) I use "topology" as a variable name in a context like this and would prefer that here.

src/fat_beagle.cpp
```diff
@@ -43,15 +43,16 @@ void FatBeagle::SetParameters(const EigenVectorXdRef param_vector) {
 
 // This is the "core" of the likelihood calculation, assuming that the tree is
 // bifurcating.
-double FatBeagle::LogLikelihoodInternals(const Tree &tree) const {
-  BeagleAccessories ba(beagle_instance_, rescaling_, tree);
+double FatBeagle::LogLikelihoodInternals(
+    const Node::NodePtr root, const std::vector<double> &branch_lengths) const {
+  BeagleAccessories ba(beagle_instance_, rescaling_, root);
```
matsen, line 48: ... and here.

src/fat_beagle.cpp
```diff
@@ -65,17 +66,18 @@ double FatBeagle::LogLikelihoodInternals(const Tree &tree) const {
 }
 
 double FatBeagle::LogLikelihood(const Tree &tree) const {
-  return LogLikelihoodInternals(DetrifurcateIfNeeded(tree));
+  return LogLikelihoodInternals(DetrifurcateIfNeeded(tree).Topology(),
+                                tree.BranchLengths());
```
matsen, line 70: This seems dangerous to me. If we _do_ detrifurcate a tree, then the branch lengths will be out of sync. 

It seems like we want 

```
auto detrifurcated_tree = DetrifurcateIfNeeded(tree);
return LogLikelihoodInternals(detrifurcated_tree.Topology(),
                                detrifurcated_tree.BranchLengths());
```

src/fat_beagle.cpp
```diff
@@ -65,17 +66,18 @@ double FatBeagle::LogLikelihoodInternals(const Tree &tree) const {
...
 }
 
 double FatBeagle::LogLikelihood(const RootedTree &tree) const {
-  return LogLikelihoodInternals(tree);
+  return LogLikelihoodInternals(tree.Topology(), tree.BranchLengths());
 }
 
 std::pair<double, std::vector<double>> FatBeagle::BranchGradientInternals(
-    const Tree &tree, const std::vector<double> &branch_lengths) const {
+    const Node::NodePtr root, const std::vector<double> &branch_lengths) const {
```
matsen, line 78: ... and here.

src/fat_beagle.hpp
```diff
@@ -78,9 +78,10 @@ class FatBeagle {
   std::pair<double, std::unordered_map<std::string, std::vector<double>>> Gradient(
       const Tree &in_tree) const;
 
-  double LogLikelihoodInternals(const Tree &tree) const;
+  double LogLikelihoodInternals(const Node::NodePtr root,
+                                const std::vector<double> &branch_lengths) const;
   std::pair<double, std::vector<double>> BranchGradientInternals(
-      const Tree &tree, const std::vector<double> &branch_lengths) const;
+      const Node::NodePtr root, const std::vector<double> &branch_lengths) const;
```
matsen, line 84: ... and here.

src/fat_beagle.cpp
4ment: good point!

PR #209, opened 2020-04-26
Unrooted versions of SBNInstance, Tree, and TreeCollection

matsen:


matsen:
Note that we're using polymorphism without pointers, like so:

```
#include <iostream>

struct Base {
  void f() { g(); };
  virtual void g() = 0;
};

struct Derived : public Base {
  virtual void g() { std::cout << "I'm in derived.\n"; };
};

int main()
{
  Derived derived;
  derived.f();
  return 0;
}
```

PR #217, opened 2020-05-07
Using mmap to store (lots of) partial likelihood vectors

matsen:


src/mmapped_plv.hpp
```diff
@@ -0,0 +1,54 @@
...
+// This class is to allocate a very large partial likelihood vector in virtual memory
+// and then cut it up (via Subdivide) into a vector of partial likelihood vectors.
+
+#ifndef SRC_MMAPPED_PLV_HPP_
+#define SRC_MMAPPED_PLV_HPP_
+
+#include "eigen_sugar.hpp"
+#include "mmapped_matrix.hpp"
+
+using NucleotidePLV = Eigen::Matrix<double, 4, Eigen::Dynamic, Eigen::ColMajor>;
```
junseonghwan, line 15: Should we create a constant for number of bases, 4? It seems to be used multiple places within this file and few places outside of this file as well.

src/mmapped_plv.hpp
```diff
@@ -0,0 +1,54 @@
...
+    auto entire_plv = mmapped_matrix_.Get();
+    const auto total_plv_length = entire_plv.cols();
+    Assert(total_plv_length % into_count == 0,
+           "into_count isn't a multiple of total PLV length in "
+           "MmappedNucleotidePLV::Subdivide.");
+    const size_t block_length = total_plv_length / into_count;
+    NucleotidePLVRefVector sub_plvs;
+    sub_plvs.reserve(into_count);
+    for (size_t idx = 0; idx < into_count; ++idx) {
+      sub_plvs.push_back(entire_plv.block(0, idx * block_length, 4, block_length));
```
junseonghwan, line 34: Replace with constant for 4.

src/gp_engine.cpp
```diff
@@ -4,11 +4,18 @@
...
+                   std::string mmap_file_path)
+    : site_pattern_(std::move(site_pattern)),
+      plv_count_(site_pattern_.PatternCount() + gpcsp_count),
+      mmapped_master_plv_(mmap_file_path, plv_count_ * site_pattern_.PatternCount()) {
+  Assert(plv_count_ > 0, "Zero PLV count in constructor of GPEngine.");
+  plvs_ = mmapped_master_plv_.Subdivide(plv_count_);
+  Assert(plvs_.size() == plv_count_,
+         "Didn't get the right number of PLVs out of Subdivide.");
+  Assert(
+      plvs_.back().rows() == 4 && plvs_.back().cols() == site_pattern_.PatternCount(),
```
junseonghwan, line 17: Another use of 4.

src/mmapped_plv.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+  NucleotidePLVRefVector Subdivide(size_t into_count) {
+    auto entire_plv = mmapped_matrix_.Get();
+    const auto total_plv_length = entire_plv.cols();
+    Assert(total_plv_length % into_count == 0,
+           "into_count isn't a multiple of total PLV length in "
+           "MmappedNucleotidePLV::Subdivide.");
+    const size_t block_length = total_plv_length / into_count;
+    NucleotidePLVRefVector sub_plvs;
+    sub_plvs.reserve(into_count);
```
junseonghwan, line 32: I probably don't understand how mmap works well enough to know what's going on here but won't this attempt to allocate large amounts of memory?

I was thinking more like there would be a function that requests for certain PLVs, and we just store those and return them.

PR #223, opened 2020-05-20
Implement gradients for clock models

4ment:


4ment:
I have not removed the ClockModel reference from PhyloModel yet but for now it is not being used.
The code in this pull request can handle strict clocks and clocks with 1 rate per branch, nothing in between.
However, a ClockModel object could be useful for intermediate models such as local clocks

matsen:
I approved-- merge whenever!

Issue #226, opened 2020-06-15
Some compilation issues that we may want to address ...


msuchard:
1. `libsbn` appears to rely on a pre-release of `Eigen` v3.4 and does not compile (in my hands) with v3.3.

msuchard:
2. We need some backwards compatibility to mac os 10.13 (because it still works with NVIDIA GPUs).  On 10.13, I get this compilation error:
```
_build/gp_engine.cpp:89:5: error: call to unavailable function 'visit': introduced in macOS 10.14
```

msuchard:
3. And this:
```
/Users/msuchard/.pyenv/versions/3.7.3/lib/python3.7/site-packages/pybind11/include/pybind11/detail/common.h:112:10: fatal error: 
      'Python.h' file not found
#include <Python.h>
```

matsen:
Thanks for this, @msuchard !

Regarding 1, the Eigen gets pulled down as part of the git submodule system. Do you not like that approach?

msuchard:
I love `git submodule`.   Will figure out why that didn't work for me, as I did try a `git submodule update`.

matsen:
I think I should have put `git submodule init` too. We could put 

    git submodule --init --recursive

in the docs.

4ment:
The eigen submodule points to a GitHub repo which is now deprecated. It would be good to use the new gitlab repo at some point

matsen:
Thanks, @4ment I have changed that now.

@msuchard , did you get past 3. by using conda?

msuchard:
Yep.

matsen:
OK. I think we're sorted with this then. Feel free to re-open if you don't agree.

PR #227, opened 2020-06-16
Fixes parts (2) and (3) of issue #226

msuchard:


matsen:
@msuchard I'll close this unless told otherwise. Thanks for trying!

msuchard:
This `PR` is critical if you want `libsbn` to compile on macOS < 10.14 (i.e. with real GPUs).

matsen:
OK, I think for that the flag

    env.Append(CXXFLAGS=["-D_LIBCPP_DISABLE_AVAILABILITY"])

is essential, but not the pyenv stuff. 

If that's correct, I'd prefer a PR with just that line because it doesn't look like we're going to try to support pyyenv. If you are just as happy for me to copy-paste the MacOS line into another branch I'm happy to do that too.

Just let me know!

msuchard:
Please copy and paste that one line, then close.

matsen:
Done in https://github.com/phylovi/libsbn/commit/25665e255ffdfec6c15fa2bc6eeb82ebea5264be

PR #228, opened 2020-06-16
Expose a ratio gradient for TensorFlow

### LINKED ISSUE #219: Expose a ratio gradient for TensorFlow
To enable a [TensorFlow custom gradient](https://www.tensorflow.org/api_docs/python/tf/custom_gradient) something like 

```python
# https://github.com/christiaanjs/treeflow/blob/master/treeflow/beagle.py#L48
@tf.custom_gradient
def ratio_gradient(x):
    heights = tf.numpy_function(forward_transform, [x], [tf.float32])
    def grad(dy):  # dy is gradient of something WRT node heights
        # Here we transform back to ratio-land.
        return tf.numpy_function(ratio_gradient, [dy], [tf.float32])
    return heights, grad
```

matsen:
Closes #219

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ format:
 	clang-format -i -style=file $(our_files)
 
 clean:
-	rm -rf _build
+	rm -rf _build build dist libsbn.*.so $(find . -name __pycache)
 
 # We follow C++ core guidelines by allowing passing by non-const reference.
 lint:
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Doing so will install the conda compiler toolchain, this will mess up our compil
 
 For your first build, do
 
-* `git submodule update`
+* `git submodule update --init --recursive`
 * `scons`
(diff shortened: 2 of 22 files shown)


matsen:
@4ment @christiaanjs we should have a wee chat about the API for RootedTree. It is easy to use incorrectly.

Right now, we have a bunch of interleaved variables, and then a `SetNodeHeightsViaHeightRatios` method that doesn't actually set the height ratios (!). 

It would seem better to have an API in which we just have a `SetHeightRatios` method that sets the height ratios and updates node heights too. 

Do we need public access to the internals?

matsen:
Dropped this "demo":

```python

def rooted_demo():
    tree = libsbn.RootedTree.example()
    height_ratios = np.array([1.0 / 3.5, 1.5 / 4.0, 7.0])
    tree.set_node_heights_via_height_ratios(height_ratios)
    ratio_gradient = libsbn.ratio_gradient_of_height_gradient(tree, height_ratios)
    assert isinstance(ratio_gradient, np.ndarray)
```

Issue #230, opened 2020-06-23
Gradient implementation of the Weibull site model


4ment:
Currently, the likelihood can be calculated with a discretized Weibull site model but we need to calculate the gradient wrt the parameter of the Weibull distribution.

matsen:
Closed by https://github.com/phylovi/libsbn/commit/a8a5a7117a651bb7fd50d34c021ffeceb19d1b66

PR #231, opened 2020-06-30
Gradient with Weibull site model

4ment:
see #230

### DIFF
--- a/src/fat_beagle.cpp
+++ b/src/fat_beagle.cpp
@@ -96,7 +96,7 @@ std::pair<double, std::vector<double>> FatBeagle::BranchGradientInternals(
   Eigen::Map<Eigen::RowVectorXd> mapQ(Q.data(), Q.size());
   EigenMatrixXd dQ = mapQ.replicate(category_count, 1);
   for (size_t k = 0; k < category_count; k++) {
-    dQ.block(0, k * matrix_dim, 1, matrix_dim) *= rates[k];
+    dQ.row(k) *= rates[k];
   }
   int derivative_matrix_idx = ba.node_count_ - 1;
   beagleSetDifferentialMatrix(beagle_instance_, derivative_matrix_idx, dQ.data());
--- a/src/fat_beagle.cpp
+++ b/src/fat_beagle.cpp
@@ -358,6 +358,17 @@ std::vector<double> ClockGradient(const RootedTree &tree,
   }
 }
 
+std::vector<double> DiscreteSiteModelGradient(
+    const std::vector<double> &branch_lengths,
+    const std::vector<double> &unscaled_category_gradient) {
+  size_t edge_count = branch_lengths.size() - 1;
+  double rate_gradient = 0;
+  for (size_t node_id = 0; node_id < edge_count; node_id++) {
+    rate_gradient += unscaled_category_gradient[node_id] * branch_lengths[node_id];
+  }
+  return {rate_gradient};
+}
+
 UnrootedPhyloGradient FatBeagle::Gradient(const UnrootedTree &in_tree) const {
   auto tree = in_tree.Detrifurcate();
   tree.SlideRootPosition();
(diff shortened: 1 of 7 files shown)


src/fat_beagle.cpp
```diff
@@ -79,24 +79,43 @@ double FatBeagle::LogLikelihood(const RootedTree &tree) const {
...
+  EigenMatrixXd dQ = mapQ.replicate(category_count, 1);
+  const EigenVectorXd &category_rates = site_model.GetCategoryRates();
+  for (size_t k = 0; k < category_count; k++) {
+    dQ.row(k) *= category_rates[k];
+  }
+  return dQ;
+}
+
+// Build differential matrix for site model gradient.
+EigenMatrixXd GetSiteModelDifferentialMatrices(SiteModel &site_model,
```
matsen, line 97: I have a slight preference for `Build` vs `Get` as the latter sounds like we are just pulling out something that's already there.

src/fat_beagle.cpp
```diff
@@ -511,16 +530,41 @@ std::vector<double> RatioGradient(const RootedTree &tree,
...
+      *phylo_model_->GetSiteModel(), *phylo_model_->GetSubstitutionModel());
   auto [log_likelihood, branch_length_gradient] =
-      BranchGradientInternals(tree.Topology(), tree.BranchLengths());
-  // We want the fixed node to have a zero gradient.
-  branch_length_gradient[tree.Topology()->Children()[1]->Id()] = 0.;
+      BranchGradientInternals(tree.Topology(), tree.BranchLengths(), dQ);
 
   std::vector<double> substitution_model_gradient;
   std::vector<double> site_model_gradient;
+  auto site_model = phylo_model_->GetSiteModel();
```
matsen, line 554: This block and the below can get factored out to reduce code duplication as well, right?

src/fat_beagle.hpp
```diff
@@ -77,7 +77,8 @@ class FatBeagle {
   double LogLikelihoodInternals(const Node::NodePtr topology,
                                 const std::vector<double> &branch_lengths) const;
   std::pair<double, std::vector<double>> BranchGradientInternals(
-      const Node::NodePtr topology, const std::vector<double> &branch_lengths) const;
+      const Node::NodePtr topology, const std::vector<double> &branch_lengths,
```
matsen, line 80: It'd be nice to have a short comment describing what dQ is.

src/rooted_sbn_instance.hpp
```diff
@@ -213,6 +213,31 @@ TEST_CASE("RootedSBNInstance: clock gradients") {
...
+  for (auto& tree : inst.tree_collection_.trees_) {
+    tree.rates_.assign(tree.rates_.size(), 0.001);
+  }
+  auto param_block_map = inst.GetPhyloModelParamBlockMap();
+  param_block_map.at(WeibullSiteModel::shape_key_).setConstant(0.1);
+
+  auto likelihood = inst.LogLikelihoods();
+  double physher_ll = -4618.2062529058;
+  CHECK_LT(fabs(likelihood[0] - physher_ll), 0.0001);
+  std::cout << likelihood[0] << std::endl;
```
matsen, line 232: Trim this output?

src/site_model.cpp
```diff
@@ -34,16 +34,29 @@ void WeibullSiteModel::SetParameters(const EigenVectorXdRef param_vector) {
...
+  double mean_rate = 0;
+  double mean_rate_derivative = 0;
+  std::vector<double> deriv_unscaled_rates(category_count_);
   for (size_t i = 0; i < category_count_; i++) {
     double quantile = (2.0 * i + 1.0) / (2.0 * category_count_);
     // Set rate to inverse CDF at quantile.
     category_rates_[i] = pow(-std::log(1.0 - quantile), 1.0 / shape_);
-    mean += category_rates_[i];
+    mean_rate += category_rates_[i];
+    // Derivative of unormalized rate i wrt shape
```
matsen, line 45: This is awesome! Do you have any helpful links for understanding the derivation?

4ment: It is coming from physher and there is even less comments in the code

PR #232, opened 2020-07-07
216 dag implementation

junseonghwan:
Implementation of generalized pruning algorithm using gradient ascent for branch optimization followed by SBN parameter optimization.

Gradient step size needs to be tuned in order to not overshoot but it's not implemented as part of this PR.

### DIFF
--- a/src/gp_doctest.cpp
+++ b/src/gp_doctest.cpp
@@ -6,8 +6,11 @@
 using namespace GPOperations;
 
 // GPCSP stands for generalized PCSP-- see text.
+
 // Let the "venus" node be the common ancestor of mars and saturn.
 enum HelloGPCSP { jupiter, mars, saturn, venus, root };
+
+size_t root_jupiter_idx = 3;
 
 // Our tree is
 // (jupiter:0.113,(mars:0.15,saturn:0.1)venus:0.22):0.;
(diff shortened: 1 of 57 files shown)


src/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# code relating to tree structured stick breaking process
```
matsen, line 1: Did you mean to include this in your PR?

src/dag_node.cpp
```diff
@@ -0,0 +1,23 @@
+// Copyright 2019-2020 libsbn project contributors.
+// libsbn is free software under the GPLv3; see LICENSE file for details.
+
+#include "dag_node.hpp"
+
+std::string GetNeighborString(std::vector<size_t> neighbors)
```
matsen, line 6: This is probably not worth fiddling with now, but have you tried using the prettyprint.hpp stuff? You can take stl containers and just << them to cout. I expect that you could << them to an ostringstream too.

src/gp_doctest.cpp
```diff
@@ -7,7 +7,7 @@ using namespace GPOperations;
 
 // GPCSP stands for generalized PCSP-- see text.
 // Let the "venus" node be the common ancestor of mars and saturn.
-enum HelloGPCSP { jupiter, mars, saturn, venus, root };
+enum HelloGPCSP { jupiter, mars, saturn, root, venus };
```
matsen, line 10: This order surprises me. Shouldn't the root have the largest ID?

src/gp_doctest.cpp
```diff
@@ -18,191 +18,78 @@ GPInstance MakeHelloGPInstance() {
...
-  EigenVectorXd branch_lengths(4);
+  EigenVectorXd branch_lengths(5);
   // Order set by HelloGPCSP.
-  branch_lengths << 0.113, 0.15, 0.1, 0.22;
+  branch_lengths << 0, 0.113, 0.22, 0.1, 0.15;
   inst.GetEngine()->SetBranchLengths(branch_lengths);
   return inst;
 }
 
-TEST_CASE("GPInstance: straightforward classical likelihood calculation") {
```
matsen, line 28: It seems like we're losing a teaching tool by getting rid of this test. Would it be hard to update to bring it into your framework?

src/gp_doctest.cpp
```diff
@@ -18,191 +18,78 @@ GPInstance MakeHelloGPInstance() {
...
-    r_t,           // PLV pointing towards venus from root
-    rhat_s,        // evolved message coming toward venus
-    r_stilde,      // PLV pointing towards mars from venus
-    r_s,           // PLV pointing towards saturn from venus
-    rhat_jupiter,  // evolved message coming towards jupiter
-    rhat_mars,     // evolved message coming towards mars
-    rhat_saturn,   // evolved message coming towards saturn
-  };
 
-  GPOperationVector two_pass_likelihood_computation{
```
matsen, line 90: Same here.

src/gp_doctest.cpp
```diff
@@ -18,191 +18,78 @@ GPInstance MakeHelloGPInstance() {
...
+  size_t leafward_idx = GetPlvIndex(PlvType::P,
+                                    hello_node_count,
+                                    child_idx);
+  size_t rootward_idx = GetPlvIndex(PlvType::R,
+                                    hello_node_count,
+                                    root_idx);
+  size_t pcsp_idx = inst.GetPCSPIndex(root_idx, child_idx, false);
+  OptimizeBranchLength op{leafward_idx, rootward_idx, pcsp_idx};
+  DoublePair log_lik_and_derivative = engine->LogLikelihoodAndDerivative(op);
+  // Expect log lik: -4.806671945.
```
matsen, line 91: Expected based on what?

src/gp_engine.cpp
```diff
@@ -8,25 +8,54 @@ GPEngine::GPEngine(SitePattern site_pattern, size_t gpcsp_count,
...
 
   auto weights = site_pattern_.GetWeights();
   site_pattern_weights_ =
-      Eigen::Map<Eigen::VectorXd, Eigen::Unaligned>(weights.data(), weights.size());
+    Eigen::Map<Eigen::VectorXd, Eigen::Unaligned>(weights.data(), weights.size());
 
   InitializePLVsWithSitePatterns();
 }
 
+GPEngine::GPEngine(SitePattern site_pattern,
```
matsen, line 31: Why do we have these two very similar constructors?

src/gp_engine.cpp
```diff
@@ -49,39 +88,31 @@ void GPEngine::operator()(const GPOperations::Multiply& op) {
...
-            << std::endl;
-}
-
-void GPEngine::operator()(const GPOperations::OptimizeLeafward& op) {
-  Failwith("OptimizeRootward unimplemented for now.");
 }
 
 void GPEngine::operator()(const GPOperations::UpdateSBNProbabilities& op) {
-  Failwith("UpdateSBNProbabilities unimplemented for now.");
+  //Failwith("UpdateSBNProbabilities unimplemented for now.");
```
matsen, line 104: delete please

src/gp_engine.cpp
```diff
@@ -49,39 +88,31 @@ void GPEngine::operator()(const GPOperations::Multiply& op) {
...
-  SetTransitionMatrixToHaveBranchLengthAndTranspose(
-      branch_lengths_(op.branch_length_idx));
-  plvs_.at(op.dest_idx) = transition_matrix_ * plvs_.at(op.src_idx);
-}
-
-void GPEngine::operator()(const GPOperations::OptimizeRootward& op) {
-  auto starting_branch_length = branch_lengths_(op.branch_length_idx);
-  std::cout << "starting branch length: " << starting_branch_length << std::endl;
+void GPEngine::operator()(const GPOperations::OptimizeBranchLength& op) {
+  //BrentOptimization(op);
```
matsen, line 99: delete please

src/gp_engine.cpp
```diff
@@ -49,39 +88,31 @@ void GPEngine::operator()(const GPOperations::Multiply& op) {
...
 void GPEngine::operator()(const GPOperations::UpdateSBNProbabilities& op) {
-  Failwith("UpdateSBNProbabilities unimplemented for now.");
+  //Failwith("UpdateSBNProbabilities unimplemented for now.");
+  size_t range_length = op.stop_idx - op.start_idx;
+  if (range_length == 1) {
+    q_(op.start_idx) = 1.;
+    return;
+  }
+
+  // Assumption: log_likelihoods_ have been updated on [op.start_idx, op.stop_idx).
```
matsen, line 111: Let's put this documentation somewhere more prominent.

src/gp_engine.cpp
```diff
@@ -100,11 +131,13 @@ void GPEngine::SetTransitionAndDerivativeMatricesToHaveBranchLength(
   Eigen::Vector4d diagonal_vector = (branch_length * eigenvalues_).array().exp();
   diagonal_matrix_.diagonal() = diagonal_vector;
   transition_matrix_ = eigenmatrix_ * diagonal_matrix_ * inverse_eigenmatrix_;
-  diagonal_matrix_.diagonal() = diagonal_vector.array() * eigenvalues_.array();
+  diagonal_matrix_.diagonal() = eigenvalues_.array() * diagonal_vector.array();
```
matsen, line 134: Interesting. I'm not sure if I understand the difference here.

src/gp_engine.cpp
```diff
@@ -150,28 +191,38 @@ void GPEngine::InitializePLVsWithSitePatterns() {
...
+void GPEngine::BrentOptimization(const GPOperations::OptimizeBranchLength& op) {
   auto negative_log_likelihood = [this, &op](double branch_length) {
     SetTransitionMatrixToHaveBranchLength(branch_length);
-    plvs_.at(op.dest_idx) = transition_matrix_ * plvs_.at(op.leafward_idx);
-    return -LogLikelihood(op.rootward_idx, op.dest_idx);
+    auto result = plvs_.at(op.rootward_idx).transpose() *
+                                    (transition_matrix_ * plvs_.at(op.leafward_idx));
+    per_pattern_log_likelihoods_ = result.diagonal().array().log();
+    double ret = -(log(q_[op.pcsp_idx]) + per_pattern_log_likelihoods_.dot(site_pattern_weights_));
+    return ret;
```
matsen, line 201: unnecessary variable

src/gp_engine.cpp
```diff
@@ -150,28 +191,38 @@ void GPEngine::InitializePLVsWithSitePatterns() {
...
+  double current_branch_length = branch_lengths_(op.pcsp_idx);
+  double current_value = negative_log_likelihood(current_branch_length);
   auto [branch_length, neg_log_likelihood] = Optimization::BrentMinimize(
       negative_log_likelihood, min_branch_length_, max_branch_length_,
       significant_digits_for_optimization_, max_iter_for_optimization_);
-  branch_lengths_(op.branch_length_idx) = branch_length;
-  log_likelihoods_(op.branch_length_idx) = -neg_log_likelihood;
+
+  // Numerical optimization sometimes yields new nllk > current nllk.
+  // In this case, we reset the branch length to previous values.
```
matsen, line 210: "to the previous value"

src/gp_engine.hpp
```diff
@@ -35,28 +39,50 @@ class GPEngine {
   void SetTransitionAndDerivativeMatricesToHaveBranchLength(double branch_length);
   void SetTransitionMatrixToHaveBranchLengthAndTranspose(double branch_length);
   const Eigen::Matrix4d& GetTransitionMatrix() { return transition_matrix_; };
+  NucleotidePLV GetPLV(size_t idx) { return plvs_[idx]; }
```
matsen, line 42: Is this used? Could it be const?

src/gp_engine.hpp
```diff
@@ -81,23 +107,23 @@ class GPEngine {
...
-  void GradientAscentOptimization(const GPOperations::OptimizeRootward& op);
+  void BrentOptimization(const GPOperations::OptimizeBranchLength& op);
+  void GradientAscentOptimization(const GPOperations::OptimizeBranchLength& op);
 
-  inline double LogLikelihood(size_t src1_idx, size_t src2_idx) {
-    per_pattern_log_likelihoods_ =
-        (plvs_.at(src1_idx).transpose() * plvs_.at(src2_idx)).diagonal().array().log();
-    return per_pattern_log_likelihoods_.dot(site_pattern_weights_);
-  }
+//  inline double LogLikelihood(size_t src1_idx, size_t src2_idx) {
```
matsen, line 113: cut?

src/dag_node.hpp
```diff
@@ -0,0 +1,57 @@
...
+  size_t id_;
+  Bitset subsplit_;
+  
+  std::vector<size_t> leafward_rotated;
+  std::vector<size_t> leafward_sorted;
+  std::vector<size_t> rootward_rotated;
+  std::vector<size_t> rootward_sorted;
+};
+
+#endif /* SRC_DAG_NODE_HPP_ */
```
matsen, line 57: The other header files use `//` here.

src/gp_instance.cpp
```diff
@@ -37,11 +58,34 @@ void GPInstance::MakeEngine() {
...
+  leafward_order_ = LeafwardPassTraversal();
+
+  InitializeGPEngine();
+}
+
+void GPInstance::MakeGPEngine() {
+  SitePattern site_pattern(alignment_, tree_collection_.TagTaxonMap());
+
+  // Get number of parameters involving fake subsplits.
+  size_t n_fake_subsplit_params = 0;
```
matsen, line 77: `_count` rather than `n_`, please.

src/gp_instance.cpp
```diff
@@ -37,11 +58,34 @@ void GPInstance::MakeEngine() {
...
+  rootward_order_ = RootwardPassTraversal();
+  leafward_order_ = LeafwardPassTraversal();
+
+  InitializeGPEngine();
+}
+
+void GPInstance::MakeGPEngine() {
+  SitePattern site_pattern(alignment_, tree_collection_.TagTaxonMap());
+
+  // Get number of parameters involving fake subsplits.
```
matsen, line 76: This seems like a reasonable thing to factor out into a method.

src/gp_instance.cpp
```diff
@@ -37,11 +58,34 @@ void GPInstance::MakeEngine() {
   CheckSequencesAndTreesLoaded();
   ProcessLoadedTrees();
   SitePattern site_pattern(alignment_, tree_collection_.TagTaxonMap());
-  engine_ = std::make_unique<GPEngine>(
-      // To count GPCSSs, we add the usual suspects to the number of leaves (which are
-      // the fake PCSS).
-      site_pattern, sbn_parameters_.size() + tree_collection_.TaxonCount(),
-      mmap_file_path_);
+
+  ConstructDAG();
```
matsen, line 62: In thinking about this before I had thought that there would be a separate GPDAG class, defined in gp_dag.* files. If it would only take an hour or two, I think that would be cleaner. What do you think?

src/gp_instance.hpp
```diff
@@ -35,16 +58,70 @@ class GPInstance {
...
+  // The remaining entries - 1 are for all other subsplits.
+  // The last entry is saved for pseudoroot -- for passing stationary.
+  // The pseudoroot is a Bitset with all 1's.
+  std::unordered_map<Bitset, size_t> subsplit_to_index_;
+  std::vector<std::shared_ptr<DAGNode>> dag_nodes_;
+
+  // Build indexing scheme for q_, branch_lengths_, log_likelihoods_.
+  // PCSP to index.
+  BitsetSizeMap pcsp_indexer_;
+  // Stores range of indices for a subsplit and rotated subsplit.
```
matsen, line 81: So the return value of this map is the concatenation of that for a subsplit and then the rotated subsplit? Given that I mentioned it before, I anticipate that there are reasons to have quite a number of different indexers that are separate than those used by the rest of libsbn, but it'd be nice to justify the difference.

src/gp_instance.hpp
```diff
@@ -35,16 +58,70 @@ class GPInstance {
...
   BitsetSizePairMap parent_to_range_;
 
+  // The first entries are reserved for fake subsplits.
+  // We will adopt the convention of denoting a trivial subsplit by placing
+  // 0's in the first chunk.
+  // The next set of entries store the rootsplits.
+  // The remaining entries - 1 are for all other subsplits.
+  // The last entry is saved for pseudoroot -- for passing stationary.
+  // The pseudoroot is a Bitset with all 1's.
+  std::unordered_map<Bitset, size_t> subsplit_to_index_;
```
matsen, line 75: There are different indices going on here, right? The PLVs have a different indexing system than the others? Would it be possible to explain that in the code?

src/gp_instance.cpp
```diff
@@ -86,3 +130,805 @@ void GPInstance::ProcessLoadedTrees() {
...
+{
+  if (!subsplit_to_index_.count(subsplit)) {
+    size_t id = dag_nodes_.size();
+    subsplit_to_index_[subsplit] = id;
+    dag_nodes_.push_back(std::make_shared<DAGNode>(id, subsplit));
+  }
+}
+
+void GPInstance::AddChildrenSubsplits(const Bitset &subsplit,
+                                      std::deque<Bitset> &q,
```
matsen, line 144: Especially given that `q` appears with a totally different meaning elsewhere, could we use a more descriptive name for it?

src/gp_instance.cpp
```diff
@@ -86,3 +130,805 @@ void GPInstance::ProcessLoadedTrees() {
...
+  }
+  GetEngine()->SetSBNParameters(q);
+
+  // Set rhat(s) = stationary for the rootsplits s.
+  std::vector<GPOperation> operations;
+  for (size_t i = 0; i < rootsplits_.size(); i++) {
+    auto rootsplit = rootsplits_[i];
+    auto root_idx = subsplit_to_index_[rootsplit + ~rootsplit];
+    operations.push_back(SetToStationaryDistribution{
+      GetPlvIndex(PlvType::R_HAT, dag_nodes_.size(), root_idx),
```
matsen, line 391: May I request "PLV" to agree with other such abbreviations in the codebase, e.g. "SBN", and even related code such as "PopulatePLVs"?

src/gp_instance.cpp
```diff
@@ -86,3 +130,805 @@ void GPInstance::ProcessLoadedTrees() {
...
+    GetEngine()->ProcessOperations(marginal_lik_operations);
+    double marginal_log_lik = GetEngine()->GetLogMarginalLikelihood();
+    std::cout << "Current marginal log likelihood: ";
+    std::cout << std::setprecision(9) << current_marginal_log_lik << std::endl;
+    std::cout << "New marginal log likelihood: ";
+    std::cout << std::setprecision(9) << marginal_log_lik << std::endl;
+    if (marginal_log_lik < current_marginal_log_lik) {
+      std::cout << "Marginal log likelihood decreased. Check branch optimization: adjust step size.\n";
+    }
+    if (abs(current_marginal_log_lik - marginal_log_lik) < tol) {
```
matsen, line 911: Yes, let's turn this into a test!

src/CMakeLists.txt
junseonghwan: Removing it from the index.

src/gp_doctest.cpp
junseonghwan: Manual calculation. Can include R script that I used for calculating if you want to see how I did the calculation.

src/gp_engine.cpp
junseonghwan: One of them was written by you and it is still called in the test https://github.com/phylovi/libsbn/blob/ee74caeb62a570f739657be4691892543b63d3dc/src/gp_engine.hpp#L148.

Should I remove the constructor even if it means to call the constructor without correct arguments?

src/gp_engine.cpp
junseonghwan: Done.

src/gp_engine.cpp
junseonghwan: Done.

src/gp_engine.cpp
junseonghwan: Any suggestion on the best place for it to go?

src/gp_engine.cpp
junseonghwan: No difference! It's just my obsession... in my derivation, I put the eigenvalues vector first. Happy to change it back.

src/gp_engine.cpp
junseonghwan: Done.

src/gp_engine.cpp
junseonghwan: Done.

src/gp_engine.hpp
junseonghwan: Not used. I put it there to test code but it's no longer needed. Removed it.

src/gp_engine.hpp
junseonghwan: Done.

src/dag_node.hpp
junseonghwan: Done.

src/gp_instance.cpp
junseonghwan: Done.

src/gp_instance.cpp
junseonghwan: Done. Changed it from `q` - > `subsplit_queue`.

src/gp_instance.cpp
junseonghwan: Done.

src/gp_instance.cpp
junseonghwan: Sorry, I'm not quite sure what you mean here.

src/gp_doctest.cpp
junseonghwan: I agree, I debated about this with myself few times as it seemed unnatural. If you also think it's unnatural, then I will change it. But this, together with bringing back old tests will take a bit more time.

junseonghwan: Done.

src/gp_instance.cpp
junseonghwan: Factor out `SitePattern site_pattern(alignment_, tree_collection_.TagTaxonMap());`??

src/gp_engine.cpp
junseonghwan: Done.

matsen:
Some deleted code:

```cpp
size_t GPDAG::GetPCSPIndex(size_t parent_node_idx, size_t child_node_idx,                                                                                                                                                         
                           bool rotated) {                                                                                                                                                                                        
  auto parent_node = dag_nodes_[parent_node_idx];                                                                                                                                                                                 
  auto child_node = dag_nodes_[child_node_idx];                                                                                                                                                                                   
  auto pcsp = rotated                                                                                                                                                                                                             
                  ? parent_node->GetBitset().RotateSubsplit() + child_node->GetBitset()                                                                                                                                           
                  : parent_node->GetBitset() + child_node->GetBitset();                                                                                                                                                           
  return pcsp_indexer_[pcsp];                                                                                                                                                                                                     
}                                                                                                                                                                                                                                 
```

PR #251, opened 2020-07-22
Make alignment compression faster

### LINKED ISSUE #238: Alignment-related operations are slow
And @ognian- has already figured out that it's because of unneeded string copying.

matsen:
Closes #238

### DIFF
--- a/src/alignment.hpp
+++ b/src/alignment.hpp
@@ -21,7 +21,7 @@ class Alignment {
 
   // Is the alignment non-empty and do all sequences have the same length?
   bool IsValid() const;
-  std::string at(const std::string& taxon) const;
+  const std::string& at(const std::string& taxon) const;
 
   static Alignment ReadFasta(const std::string& fname);
 
--- a/src/site_pattern.hpp
+++ b/src/site_pattern.hpp
@@ -15,8 +15,8 @@
 class SitePattern {
  public:
   SitePattern() = default;
-  explicit SitePattern(const Alignment& alignment, const TagStringMap& tag_taxon_map)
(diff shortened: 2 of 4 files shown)


src/alignment.hpp
```diff
@@ -21,7 +21,7 @@ class Alignment {
 
   // Is the alignment non-empty and do all sequences have the same length?
   bool IsValid() const;
-  std::string at(const std::string& taxon) const;
+  const std::string& at(const std::string& taxon) const;
```
ognian-, line 24: The same change should be made to the definition in alignment.cpp

matsen: 🤦 fixed

src/site_pattern.hpp
```diff
@@ -15,8 +15,8 @@
 class SitePattern {
  public:
   SitePattern() = default;
-  explicit SitePattern(const Alignment& alignment, const TagStringMap& tag_taxon_map)
-      : alignment_(alignment), tag_taxon_map_(tag_taxon_map) {
+  SitePattern(const Alignment& alignment, TagStringMap tag_taxon_map)
```
ognian-, line 18: Better change tag_taxon_map to be rvalue (TagStringMap&& tag_taxon_map), since it will be moved from.

matsen: OK, I'm glad to have you to talk this over. See 

https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference

ognian-: You're right - my comment is not valid. I assumed the constructor is meant to only accept rvalues, but it's used with lvalues in the rest of the code.

PR #256, opened 2020-07-23
Rescale when maximum drops below a threshold

matsen:
It turns out that the issue was that we were rescaling too much, and we were _overflowing_. I tried a few things and the safest seems to just make sure that the max entry doesn't exceed 1-epsilon.


matsen:
You'll also note that I renamed `plv_rescaling_counts_` to just `rescaling_counts_`... this is the only rescaling we are doing so it's unambiguous and "rescaling" without plv already appeared elsewhere.

PR #260, opened 2020-08-04
234 uniform topology

junseonghwan:
Features added:
- Implements uniform prior over the tree topologies.
- Generate all trees spanned by the subsplits (DAG).

Input requested: lacking in the way of tests, how should we go about testing this code?

### DIFF
--- a/src/gp_dag.cpp
+++ b/src/gp_dag.cpp
@@ -37,12 +37,12 @@ void GPDAG::CountTopologies() {
       }
     }
   }
-  topology_count = 0;
+  topology_count_ = 0;
   IterateOverRootsplitIds(
-      [this](size_t root_id) { topology_count += topology_count_below_[root_id]; });
+      [this](size_t root_id) { topology_count_ += topology_count_below_[root_id]; });
 }
 
-double GPDAG::TopologyCount() const { return topology_count; }
+double GPDAG::TopologyCount() const { return topology_count_; }
 
 size_t GPDAG::NodeCount() const { return dag_nodes_.size(); }
 
@@ -119,28 +119,28 @@ EigenVectorXd GPDAG::BuildUniformQ() const {
 Node::NodePtrVec GPDAG::GenerateAllGPNodeIndexedTopologies() const {
   std::vector<Node::NodePtrVec> topology_below(NodeCount());
 
-  auto GetSubtrees = [&topology_below](const GPDAGNode *node) {
-    Node::NodePtrVec rotated_subtrees, sorted_subtrees;
+  auto GetSubtopologies = [&topology_below](const GPDAGNode *node) {
+    Node::NodePtrVec rotated_subtopologies, sorted_subtopologies;
     for (const bool rotated : {false, true}) {
       for (const auto &child_id : node->GetLeafward(rotated)) {
-        for (const auto &sub_tree : topology_below.at(child_id)) {
-          rotated ? rotated_subtrees.push_back(sub_tree)
-                  : sorted_subtrees.push_back(sub_tree);
+        for (const auto &subtopology : topology_below.at(child_id)) {
+          rotated ? rotated_subtopologies.push_back(subtopology)
+                  : sorted_subtopologies.push_back(subtopology);
         }
       }
     }
-    return std::make_pair(rotated_subtrees, sorted_subtrees);
+    return std::make_pair(rotated_subtopologies, sorted_subtopologies);
   };
 
-  auto MergeTrees = [](size_t node_id, Node::NodePtrVec &rotated_subtrees,
-                       Node::NodePtrVec &sorted_subtrees) {
+  auto MergeTopologies = [](size_t node_id, Node::NodePtrVec &rotated_subtopologies,
+                            Node::NodePtrVec &sorted_subtopologies) {
     Node::NodePtrVec topologies;
-    for (size_t i = 0; i < rotated_subtrees.size(); i++) {
-      auto &rotated_subtree = rotated_subtrees.at(i);
-      for (size_t j = 0; j < sorted_subtrees.size(); j++) {
-        auto &sorted_subtree = sorted_subtrees.at(j);
+    for (size_t i = 0; i < rotated_subtopologies.size(); i++) {
+      auto &rotated_subtopology = rotated_subtopologies.at(i);
+      for (size_t j = 0; j < sorted_subtopologies.size(); j++) {
+        auto &sorted_subtopology = sorted_subtopologies.at(j);
         Node::NodePtr new_topology =
-            Node::Join(sorted_subtree, rotated_subtree, node_id);
+            Node::Join(sorted_subtopology, rotated_subtopology, node_id);
         topologies.push_back(new_topology);
       }
     }
@@ -152,8 +152,9 @@ Node::NodePtrVec GPDAG::GenerateAllGPNodeIndexedTopologies() const {
     if (node->IsLeaf()) {
       topology_below.at(node_id).push_back(Node::Leaf(node_id));
     } else {
-      auto [rotated_trees, sorted_trees] = GetSubtrees(node);
-      topology_below[node_id] = MergeTrees(node_id, rotated_trees, sorted_trees);
+      auto [rotated_topologies, sorted_topologies] = GetSubtopologies(node);
+      topology_below[node_id] =
+          MergeTopologies(node_id, rotated_topologies, sorted_topologies);
     }
   }
 
@@ -164,7 +165,7 @@ Node::NodePtrVec GPDAG::GenerateAllGPNodeIndexedTopologies() const {
   });
 
   Assert(topologies.size() == TopologyCount(),
-         "The realized number of trees does not match the expected count.");
+         "The realized number of topologies does not match the expected count.");
 
   return topologies;
 }
@@ -192,7 +193,7 @@ EigenVectorXd GPDAG::BuildUniformPrior() const {
   IterateOverRootsplitIds([this, &q](size_t root_id) {
     auto node = GetDagNode(root_id);
     auto gpcsp_idx = gpcsp_indexer_.at(node->GetBitset());
-    q[gpcsp_idx] = topology_count_below_[root_id] / topology_count;
+    q[gpcsp_idx] = topology_count_below_[root_id] / topology_count_;
   });
 
   return q;
--- a/src/gp_dag.cpp
+++ b/src/gp_dag.cpp
@@ -11,8 +11,10 @@ using namespace GPOperations;
 
 GPDAG::GPDAG() : taxon_count_(0), rootsplit_and_pcsp_count_(0) {}
 
-GPDAG::GPDAG(const RootedTreeCollection &tree_collection) {
-  ProcessTrees(tree_collection);
+GPDAG::GPDAG(const RootedTreeCollection &tree_collection) :
+tree_collection_(tree_collection)
+{
+  ProcessTrees(tree_collection_);
   BuildNodes();
   BuildEdges();
   SetPCSPIndexerEncodingToFullSubsplits();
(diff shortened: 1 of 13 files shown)


src/gp_dag.cpp
```diff
@@ -106,6 +109,87 @@ EigenVectorXd GPDAG::BuildUniformQ() const {
   return q;
 }
 
+Node::NodePtrVec GPDAG::GenerateAllTrees() const {
+  std::vector<Node::NodePtrVec> trees_below(NodeCount());
+
+  auto GetSubtrees = [&trees_below](const GPDAGNode *node,
+                                          Node::NodePtrVec &rotated_subtrees,
+                                          Node::NodePtrVec &ordered_subtrees) {
```
matsen, line 117: I believe we are using "sorted" as the opposite as "rotated", not "ordered".

src/gp_dag.cpp
```diff
@@ -106,6 +109,87 @@ EigenVectorXd GPDAG::BuildUniformQ() const {
...
+    }
+  };
+
+  for (const auto &node_id : RootwardPassTraversal()) {
+    const auto &node = GetDagNode(node_id);
+    if (node->IsLeaf()) {
+      trees_below.at(node_id).push_back(Node::Leaf(node_id));
+    } else {
+      Node::NodePtrVec rotated_nodes, ordered_nodes;
+      GetSubtrees(node, rotated_nodes, ordered_nodes);
```
matsen, line 148: Could you please refactor these functions so that they return vectors rather than use output arguments? I realize that we do frequently use this pattern when we need to modify something, but in general returning outputs is preferred.

src/gp_dag.hpp
```diff
@@ -44,9 +47,23 @@ class GPDAG {
   // Get the index of a PLV of a given type and with a given index.
   static size_t GetPLVIndexStatic(PLVType plv_type, size_t node_count, size_t src_idx);
   size_t GetPLVIndex(PLVType plv_type, size_t src_idx) const;
+  
+  size_t GetGPCSPIndex(const Bitset &pcsp) const {
```
matsen, line 51: Could we rename this "GetGPCSPIndexWithDefault" to clearly signal that it has special behavior?

Also, can we move this method and the method below into a cpp file?

src/gp_instance.cpp
```diff
@@ -174,3 +175,63 @@ void GPInstance::EstimateSBNParameters() {
   double marginal_log_lik = GetEngine()->GetLogMarginalLikelihood();
   std::cout << std::setprecision(9) << marginal_log_lik << std::endl;
 }
+
+RootedTreeCollection GPInstance::GetRootedTreeCollection() {
```
matsen, line 179: I suggest "GenerateCompleteRootedTreeCollection" or "GenerateAllTrees". In particular, I like to reserve "Get" for accessors.

src/gp_instance.cpp
```diff
@@ -174,3 +175,63 @@ void GPInstance::EstimateSBNParameters() {
...
+
+  Tree::TreeVector tree_vector;
+  Node::NodePtrVec trees = dag_.GenerateAllTrees();
+  EigenVectorXd bl = engine_->GetBranchLengths();
```
matsen, line 183: const ref?

src/gp_instance.cpp
```diff
@@ -174,3 +175,63 @@ void GPInstance::EstimateSBNParameters() {
...
+
+  // Leaves encoding to parent subsplit encoding.
+  std::unordered_map<const Node*, Bitset> leaves2subsplit_indexer;
```
matsen, line 186: "to" rather than "2" to be consistent with the rest of the code, please.

src/gp_instance.cpp
```diff
@@ -174,3 +175,63 @@ void GPInstance::EstimateSBNParameters() {
...
+    {
+      GPDAGNode *dag_node = dag_.GetDagNode(node->Id());
+      if (!leaves2subsplit_indexer.count(node)) {
+        SafeInsert(leaves2subsplit_indexer, node, dag_node->GetBitset());
+      }
+
+      for (size_t i = 0; i < node->Children().size(); i++) {
+        size_t child_id = node->Children().at(i)->Id();
+        GPDAGNode *child_node = dag_.GetDagNode(child_id);
+        size_t i0 = dag_.GetGPCSPIndex(dag_node->GetBitset(false) + child_node->GetBitset());
```
matsen, line 198: What's interesting about this is that you've added a default for GetGPCSPIndex, but then as far as I can tell the main purpose of the default is to raise an error if it's missing? Why not just raise an error if it's out of bounds in the GetGPCSPIndex function?

src/gp_instance.cpp
```diff
@@ -174,3 +175,63 @@ void GPInstance::EstimateSBNParameters() {
...
+        size_t i1 = dag_.GetGPCSPIndex(dag_node->GetBitset(true) + child_node->GetBitset());
+        size_t gpcsp_idx = std::min(i0, i1);
+        Assert(gpcsp_idx < SIZE_MAX, "GPCSP does not exist.");
```
matsen, line 201: So the point of this block of code is to make sure that everything exists?

src/gp_dag.cpp
```diff
@@ -106,6 +109,87 @@ EigenVectorXd GPDAG::BuildUniformQ() const {
   return q;
 }
 
+Node::NodePtrVec GPDAG::GenerateAllTrees() const {
```
matsen, line 112: I realize I wasn't consistent about this, but I'd prefer to substitute every instance of "tree" in this file and with "topology". That's more accurate and is what I've tried to do elsewhere.

src/gp_dag.cpp
```diff
@@ -106,6 +109,87 @@ EigenVectorXd GPDAG::BuildUniformQ() const {
...
+  }
+
+  Node::NodePtrVec trees;
+  IterateOverRootsplitIds([&trees, &trees_below](size_t root_id) {
+    trees.insert(trees.end(), trees_below.at(root_id).begin(), trees_below.at(root_id).end());
+  });
+  
+  Assert(trees.size() == TreeCount(), "The realized number of trees does not match the expected count.");
+
+  return trees;
```
matsen, line 161: Let's call `Polish` on all the trees here.

src/gp_instance.cpp
```diff
@@ -174,3 +175,63 @@ void GPInstance::EstimateSBNParameters() {
...
+      }
+    });
+  }
+
+  for (auto &root_node : trees) {
+    // Polish will re-assign the node Ids.
+    root_node->Polish();
```
matsen, line 208: As suggested above, I suggest polishing as part of the topology construction process.

src/gp_instance.cpp
junseonghwan: Removed this block of code.

src/gp_dag.cpp
junseonghwan: Done.

src/gp_dag.hpp
junseonghwan: Done.

src/gp_dag.cpp
junseonghwan: Done.

src/gp_instance.cpp
junseonghwan: Done.

src/gp_instance.cpp
junseonghwan: Done.

src/gp_instance.cpp
junseonghwan: Removed.

src/gp_dag.cpp
junseonghwan: Done.

src/gp_instance.cpp
junseonghwan: If I'm not mistaken, the Node class sorts the children? So even though when I'm building the tree, I put the rotated child as the left child and the sorted child as the right child, this order is not held up later when we look up the index for PCSP.  I look up parent rotated subsplit + child subsplit and parent sorted subsplit + child subsplit and if I get the default for both, I raise an error as I expect exactly one to return a valid index.

src/gp_instance.cpp
junseonghwan: Using `EigenVectorXdRef`fails to compile. Not sure why.

src/gp_dag.cpp
junseonghwan: I assign the DAG Ids as the node IDs. And these are used by the caller of this function, in particular, `GetRootedTreeCollection` to look up the GPCSP index. Polish will reassign node IDs so we aren't ready to do that yet here.

src/gp_instance.cpp
junseonghwan: If you prefer that Polish be called right after the topology is constructed, I will have to restructure the code.  I don't have preference either way, but if you have strong preference that let me know.

Recall that `GPDAG` does not have access to the branch lengths, which are in the `GPEngine`. So the way it works now is for `GPInstance` to 
1. Get the topologies from `GPDAG::GenerateAllTrees`, The ID assigned to each Node instance upon construction is the corresponding `GPDAGNode` ID. Note: calling Polish re-assigns node IDs.
2. Get the branch lengths from `GPEngine`,
3. Traverse each topology to visit each node, use their Node IDs to get the `GPDAGNode` instance and the corresponding `Bitset` representation, store them in an indexer (`NodePtr` to `Bitset`).
4. Do a second traversal, this time look up the index for PCSP and set the branch length.

src/gp_dag.cpp
matsen: OK, this makes sense! Let's just signal that to the caller by renaming the method GenerateAllGPNodeIndexedTopologies or something, with a comment clarifying that this is what you are doing.

src/gp_instance.cpp
matsen: Got it. Let's rename to `GetPCSPIndexWithDefault` and move on. Thanks!

src/gp_dag.cpp
junseonghwan: OK. Would you like to keep the lambdas or should I factor them out as their own functions within the class?

matsen: I would prefer factoring out to functions. Thanks.

src/gp_dag.cpp
junseonghwan: Done.

src/gp_instance.cpp
junseonghwan: Done.

src/gp_dag.cpp
junseonghwan: I'll keep the lambdas but have changed the functions to return vectors.

junseonghwan:
I have made the requested changes. I included a test to ensure that the trees are unique; added a dummy fasta file `five_taxon_rooted.fast` to be used with `five_taxon_rooted.nwk` in the testing.

src/gp_instance.cpp
```diff
@@ -174,3 +175,57 @@ void GPInstance::EstimateSBNParameters() {
...
+  Node::NodePtrVec topologies = dag_.GenerateAllGPNodeIndexedTopologies();
+  const EigenVectorXd gpcsp_indexed_branch_lengths = engine_->GetBranchLengths();
+
+  // Map from Leaves-encoding to parent-subsplit encoding.
+  std::unordered_map<const Node *, Bitset> leaves_to_subsplit_indexer;
+  for (const auto &topology : topologies) {
+    topology->PreOrder([this, &leaves_to_subsplit_indexer](const Node *node) {
+      GPDAGNode *dag_node = dag_.GetDagNode(node->Id());
+      if (leaves_to_subsplit_indexer.count(node) == 0) {
+        SafeInsert(leaves_to_subsplit_indexer, node, dag_node->GetBitset());
```
matsen, line 190: You describe this as a map from a leaves encoding to a parent-subsplit encoding, but actually the mapping is from an entire Node, which has other things besides the leaves encoding. This is what you wanted?

junseonghwan: Yes. I was thinking to use node->Leaves() initially but realized that the node ptr is also unique. We could go back to node->Leaves() as key or remove the comment "leaves encoding" to "node ptr".

matsen: Storing a pointer as a key to a map feels strange to me. Change however you like and I'll have a look in the AM.

Issue #266, opened 2020-08-13
Refactor indexers into a stand-alone class


matsen:
Indexers are maps from bitset representations of clades/subsplits/PCSPs to indices where we store information about them.

We currently implement indexers as "raw" unordered maps, with helper functions being namespace-delimited free functions defined in `sbn_maps.hpp`. There are collections of indexers that go together for the rooted and unrooted cases, which are currently handled in the corresponding version of an SBN instance. This leads to some strange code, such as this "Bundle" which is only used as a return type: `using IndexerBundle = std::tuple<BitsetVector, BitsetSizeMap, SizeBitsetMap, BitsetSizePairMap, size_t>`.

The proximal motivation, though, is that we'd like to add information to these indexers that will allow us to handle them more flexibly. For example, the connection between the taxon names and the taxa in the indexers comes from the containing SBNInstance. Thus, if we want to do subsplit mutualization a la @mdkarcher then we would need more complex free functions, which would require us to keep more ad-hoc data structures in line.

We propose to build a series of Indexer classes that

* have taxon information built into the class
* wraps together the indexer_ (which is a PCSP indexer), rootsplits_, index_to_child_, parent_to_range_, and provides abstracted means of getting what we want from these data structures
* are made with an inheritance hierarchy that factors up the common code
* makes these core data structures private, and no longer accessible from Python (cc @zcrabbit)

This is going to set the foundation for merging these indexers for supertrees and online work.

I also may take on #247 and some of #236 as part of this.

PR #268, opened 2020-08-17
[Un]RootedSBNSupport class; make SBNInstance templated on TreeCollection and SBNSupport types.

matsen:
## Description

[Un]RootedSBNSupport class; make SBNInstance templated on TreeCollection and SBNSupport types.

Fixes issue #266

## Tests

Refactoring only, so no new tests added.

## Checklist:


src/generic_sbn_instance.hpp
```diff
@@ -0,0 +1,323 @@
...
+    };
+    return Node::Join(process_subsplit(parent_subsplit),
+                      process_subsplit(parent_subsplit.RotateSubsplit()));
+  }
+
+  // Clear all of the state that depends on the current tree collection.
+  void ClearTreeCollectionAssociatedState() {
+    sbn_parameters_.resize(0);
+    topology_counter_.clear();
+    sbn_support_ = TSBNSupport();
```
matsen, line 258: @ognian- style comments here? This is wiping out the previous `sbn_support_` by instead assigning a default-constructed one. This should activate the destructor, right? (I note that the "destructor" is a rule-of-zero one.)

Thank you for answering my newbie questions :)

ognian-: @matsen The destructor won't be called here, this calls the assignment operator. The destructor will be called only on the temporary object

src/mersenne_twister.hpp
```diff
@@ -0,0 +1,19 @@
...
+#include <random>
+
+class MersenneTwister {
+ public:
+  inline void SetSeed(unsigned long seed) { random_generator_.seed(seed); }
+  inline std::mt19937 &GetGenerator() const { return random_generator_; };
+
+ private:
+  static std::random_device random_device_;
+  static std::mt19937 random_generator_;
```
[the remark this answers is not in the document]
ognian-: @matsen It may be better to leave them non-static, and instantiate MerseneTwister only once with a static linkage, taking seed in the constructor

src/generic_sbn_instance.hpp
matsen: Oh!

Here is a simplified version to be more precise.

Let's say `Blah` is a class that takes a single `size_t` and makes a vector of floats out of it:

```cpp

struct Blah {
    Blah(size_t size) : my_vector_(std::vector(size)) {};
    std::vector<float> my_vector_
}

x = Blah(1)
x = Blah(2)
```
what happens to the vector from `Blah(1)` when we reassign?

src/mersenne_twister.hpp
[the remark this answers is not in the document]
matsen: Thanks @ognian- ! I'm not sure what you mean by "static linkage". Do you mean that we have a static MerseneTwister? I'd prefer not to do that, because I want this class to sit inside a templated class, which seemed to be causing troubles. 

We may want to set the seed after construction.

Do you see problems with this approach, or is it more just ugly?

src/generic_sbn_instance.hpp
ognian-: In this example vector's assignment operator is called, this happens for every member. No destructors are called until "x" goes out of scope.

src/mersenne_twister.hpp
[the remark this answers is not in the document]
ognian-: My point is mainly for ugliness, not that it wouldn't work. This has the appearance that the class has state, thus the users should instantiate it and use that instance. But setting the seed on any instance would set it for all instances. I think it would be cleaner if the class is a singleton.

Issue #235, opened 2020-07-09
Move branch length optimization into log space


matsen:

matsen:
* could try asking for more sig figs out of Brent
* could try doing brent in log space

matsen:
* Adaptive stepsize for generalized pruning gradient descent

matsen:
Scope shrunk, and everything else added to #278

PR #276, opened 2020-08-30
234 4 uniform toplogy final

junseonghwan:
## Description

+ Changes all likelihood calculations to be product-sum as we found out that GP does not compute sum-product. 
+ `GPEngine` maintains log likelihood as matrix, with the number of rows equal to number of GPCSPs and the number of columns equal to the number of sites.
+ All subsequent computation, particularly `GPEngine::UpdateSBNProbabilities` and `GPInstance::EstimateSBNParameters` code that uses the log likelihood computation has been updated.

Fixes issue #235 -- cherry picked commits to 235-optimization by @hrnasif (see 7140dec018c056c74802fa7158b7c28bcb96f479).

## Tests
All tests are in `src/gp_doctest.cpp`.
+ Wrote function for computing exact prod-sum likelihood given unrooted newick and fasta.
+ To implement above, we allowed extracting a single site from `Alignment` instance and a setter to supply `Alignment` object to `GenericSBNInstance` instance.
+ Fixed and strengthened old tests to compare against exact prod-sum likelihood.
+ Added marginal likelihood test on five taxa data involving 4 trees.
+ Ensure PLVs are populated correctly during `EstimateBranchLengths`.
+ Ensure SBN parameters are updated correctly.

## Checklist:


src/gp_doctest.cpp
```diff
@@ -191,5 +204,113 @@ TEST_CASE("GPInstance: Hotstart branch lengths") {
...
+  auto inst = MakeFiveTaxaInstance();
+  inst.EstimateBranchLengths(1e-6, 10);
+  double gp_marginal_log_likelihood = inst.GetEngine()->GetLogMarginalLikelihood();
+  double exact_marginal_log_likelihood = ComputeExactMarginal(
+      "data/five_taxon_unrooted_with_branch_lengths.nwk", "data/five_taxon.fasta");
+  CHECK_LT(abs(exact_marginal_log_likelihood - gp_marginal_log_likelihood), 1e-6);
+}
+
+TEST_CASE("GPInstance: test populate PLV") {
+  // Estimate branch lengths. This will populate and update PLVs. Then compute
```
matsen, line 241: Could this be summarized by saying "This test makes sure that PopulatePLVs correctly re-populates the PLVs using the current branch lengths?" 

The fact that there is so much explanation here and below makes me think that `PopulatePLVs` should be called `ResetMarginalLikelihoodAndPopulatePLVs`.

src/gp_doctest.cpp
```diff
@@ -191,5 +208,111 @@ TEST_CASE("GPInstance: Hotstart branch lengths") {
...
+  inst.EstimateBranchLengths(1e-6, 10);
+  inst.ComputeLikelihoods();
+  const EigenVectorXd log_likelihoods1 = inst.GetEngine()->GetLogLikelihoods();
+  inst.PopulatePLVs();
+  inst.ComputeLikelihoods();
+  const EigenVectorXd log_likelihoods2 = inst.GetEngine()->GetLogLikelihoods();
+  CheckVectorXdEquality(log_likelihoods1, log_likelihoods2, 1e-6);
+}
+
+TEST_CASE("GPInstance: SBN probabilities on five taxa") {
```
matsen, line 255: I don't want to diminish what you've done, but it seems to me that this only tests SBN probabilities at the rootsplits, right? If so, let's just add "rootsplit probabilities" or someting to the name here.

src/gp_engine.cpp
```diff
@@ -41,7 +43,8 @@ void GPEngine::operator()(const GPOperations::Zero& op) {
 void GPEngine::operator()(const GPOperations::SetToStationaryDistribution& op) {
   auto& plv = plvs_.at(op.dest_);
   for (size_t row_idx = 0; row_idx < plv.rows(); ++row_idx) {
-    plv.row(row_idx).array() = stationary_distribution_(row_idx);
+    plv.row(row_idx).array() =
```
matsen, line 46: I think I asked for a comment here about why this makes sense.

src/gp_engine.hpp
```diff
@@ -143,13 +156,15 @@ class GPEngine {
...
   }
 
-  inline void PreparePerPatternLogLikelihoods(size_t src1_idx, size_t src2_idx) {
-    per_pattern_log_likelihoods_ =
-        (plvs_.at(src1_idx).transpose() * transition_matrix_ * plvs_.at(src2_idx))
-            .diagonal()
-            .array()
-            .log() +
-        LogRescalingFor(src1_idx) + LogRescalingFor(src2_idx);
+  inline void PreparePerPatternLogLikelihoods(size_t gpcsp_idx, size_t src1_idx,
```
matsen, line 159: Could we have a corresponding comment here about what these per-pattern log likes are, now that they aren't quite standard?

src/gp_engine.hpp
```diff
@@ -92,7 +105,7 @@ class GPEngine {
   // These parameters are indexed in the same way as sbn_parameters_ in
   // gp_instance.
   EigenVectorXd branch_lengths_;
-  EigenVectorXd log_likelihoods_;
+  EigenMatrixXd log_likelihoods_;
```
matsen, line 108: I don't think that `log_likelihoods_` follows 

    // These parameters are indexed in the same way as sbn_parameters_ in
    // gp_instance.

Please move below and explain what the rows and columns of this matrix are in a comment.

src/gp_engine.hpp
```diff
@@ -67,13 +78,15 @@ class GPEngine {
  private:
   static constexpr double min_branch_length_ = 1e-6;
   static constexpr double max_branch_length_ = 3.;
+  static constexpr double min_log_branch_length_ = -13.9;
```
matsen, line 81: It doesn't seem like we use the non-log versions anymore, so let's discard them.

src/gp_engine.hpp
```diff
@@ -67,13 +78,15 @@ class GPEngine {
...
+  static constexpr double max_log_branch_length_ = 1.1;
 
   int significant_digits_for_optimization_ = 6;
   double relative_tolerance_for_optimization_ = 1e-2;
   double step_size_for_optimization_ = 5e-4;
   size_t max_iter_for_optimization_ = 1000;
 
-  double log_marginal_likelihood_ = DOUBLE_NEG_INF;
+  EigenVectorXd log_marginal_likelihood_;
```
matsen, line 89: Please describe the entries of this vector in a comment.

src/gp_instance.cpp
```diff
@@ -176,21 +179,25 @@ void GPInstance::EstimateBranchLengths(double tol, size_t max_iter) {
   std::cout << "initial likelihood: " << initial_likelihood_duration.count() << "s\n";
   std::cout << "optimization: " << optimization_duration.count() << "s or "
             << optimization_duration.count() / 60 << "m\n";
+  std::ofstream branch_length_file("_output/branch_lengths.txt");
+  branch_length_file << GetEngine()->GetBranchLengths();
```
matsen, line 183: We could come up with a `BranchLengthsToCSV` function like `SBNParametersToCSV` that also includes indexer information. Would this be useful?

src/gp_operation.hpp
```diff
@@ -35,9 +35,13 @@ struct Zero {
 
 // Set the PLV at `dest_` to be the stationary distribution at every site.
 struct SetToStationaryDistribution {
-  constexpr explicit SetToStationaryDistribution(size_t dest) : dest_{dest} {}
+  constexpr explicit SetToStationaryDistribution(size_t dest, size_t root_gpcsp_idx)
```
matsen, line 38: The comment for this needs updating.

src/gp_engine.cpp
```diff
@@ -245,38 +246,38 @@ double GPEngine::LogRescalingFor(size_t plv_idx) {
...
-    branch_lengths_(op.gpcsp_) = branch_length;
+    branch_lengths_(op.gpcsp_) = exp(log_branch_length);
   }
 }
 
 void GPEngine::GradientAscentOptimization(
     const GPOperations::OptimizeBranchLength& op) {
-  auto log_likelihood_and_derivative = [this, &op](double branch_length) {
-    branch_lengths_(op.gpcsp_) = branch_length;
+  auto log_likelihood_and_derivative = [this, &op](double log_branch_length) {
```
matsen, line 272: I would have thought that something would have needed to change with `LogLikelihoodAndDerivative` when switching to log branch lengths. Am I missing something?

data/hello_unrooted_two_trees.nwk
```diff
@@ -0,0 +1,2 @@
+(jupiter:0.0697739,mars:0.204056,saturn:0.066997):0;
+(jupiter:0.0671995,mars:0.203894,saturn:0.0695548):0;
```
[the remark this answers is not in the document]
junseonghwan: Sure, although I actually don't see any symbol...

Screen Shot 2020-08-31 at 9 19 39 PM

But it got me interested in why we always need to end a file with a new line: https://stackoverflow.com/a/729795. Didn't know that it was a POSIX standard!!

src/gp_doctest.cpp
junseonghwan: Comments updated. 

`ResetMarginalLikelihoodAndPopulatePLVs` is more informative. I've renamed `PopulatePLVs` to that.

src/gp_doctest.cpp
junseonghwan: Done.

src/gp_engine.cpp
junseonghwan: I initially put it in the header file. It's now moved here.

src/gp_engine.hpp
junseonghwan: Done.

src/gp_engine.hpp
junseonghwan: Done. `min_branch_length_` and `max_branch_length_` are removed.

src/gp_engine.hpp
junseonghwan: Done.

src/gp_operation.hpp
junseonghwan: Done.

src/gp_engine.cpp
junseonghwan: `GradientAscentOptimization` and its related functions have not been modified nor tested as part of this pull request.  

I will create a separate issue to investigate specifically for gradient based optimization of the branch lengths.

src/gp_instance.cpp
junseonghwan: Yes, I think it would be useful.

src/gp_engine.hpp
junseonghwan: I wrote some comments to indicate that these are actually computing marginal log likelihood over the trees that include the GPCSP corresponding to `gpcsp_idx`. 

Should we change the function name to reflect that? `PreparePerPatternMarginalLogLikelihoods` is probably going to cause confusion since these aren't marginal over all trees but rather a subset of the trees. `PreparePerPatternMarginalLogLikelihoodsOverGPCSPTrees`? Suggestions?

junseonghwan:
- Could you please cut out all (or at least most) of the print output in gp_doctest?
    + Done.
- It does seem like optimizing branch length in log space is a win, but is no further discussion needed?
    + Like comparing the performance in terms of log marginal likelihood when optimizing in log space vs not? Could do some simple test runs to check that optimization in log space leads to better results. Eye test shows that it is able to get finer values for the small branches lengths.
- To be clear, the changes to the fixed values in gp_doctest is because we are now doing better optimization? Not because of the switch to log branch lengths as such?
    + Yes. Previously, when the branch lengths are small, they just get fixed at some value (i.e, 0.015...). The regression test were failing because the branch length that we get with log-space optimization are no longer getting fixed at some threshold. 
    + Other tests were also failing for the same reason. It depended on files that were generated after running `EstimateBranchLength` to get the branch lengths. But with the change in the optimization, we weren't getting the same values.

src/gp_engine.cpp
matsen: I made an issue https://github.com/phylovi/libsbn/issues/278 and linked to it in the code.

src/gp_engine.hpp
```diff
@@ -47,10 +47,17 @@ class GPEngine {
...
-  double GetLogMarginalLikelihood() const { return log_marginal_likelihood_; }
+  void ResetLogMarginalLikelihood() {
+    log_marginal_likelihood_.setConstant(DOUBLE_NEG_INF);
+  }
+  double GetLogMarginalLikelihood() const {
+    return (log_marginal_likelihood_.array() * site_pattern_weights_.array()).sum();
+  }
   EigenVectorXd GetBranchLengths() const { return branch_lengths_; };
-  EigenVectorXd GetLogLikelihoods() const { return log_likelihoods_; };
+  EigenVectorXd GetLogLikelihoods() const {
```
matsen, line 57: Thanks for documenting `log_likelihoods_`. Could you please put some documentation about what this function returns? 

I think that it's a vector indexed by GPCSPs, such that entry i stores the marginal log likelihood over all trees that include a GPCSP indexed by i.

src/gp_engine.hpp
matsen: I would suggest `PreparePerPatternLogLikelihoodsForGPCSP`. Thanks!

src/gp_engine.hpp
matsen: Also, given that we have a `log_likelihoods_` matrix and this doesn't return this, I might suggest `GetPerGPCSPLogLikelihoods` or something.

src/gp_engine.hpp
junseonghwan: Done.

src/gp_engine.hpp
junseonghwan: Done.

Issue #278, opened 2020-09-01
Further development and testing of GP optimization


matsen:
Moving branch lengths to log space broke our gradient-based optimization, so we need to update and test that.

We should also better understand the properties of the global optimization problem and our coordinate-based approach, as well as benchmark gradient descent vs Brent.

matsen:
I'm also curious how frequently the optimization procedure decreases the likelihood function that it's meant to optimize (we have some code catching that).

PR #279, opened 2020-09-03
More flexible tree parsing and time tree handling

### LINKED ISSUE #277: Tree parsing issues
Some issues I have noticed:

Programs include comments inside a newick tree file in different ways. Currently libsbn parses `(A:[&date=1]0.1, B:1)` correctly but it does not work on `(A[&date=1]:0.1, B:1)`

In some cases we can provide a topology for the purpose of estimating a time tree. Libsbn expects the branches to represent time (e.g. ultrametric) so it will fail if the time constraints are not satisfied. Can we remove this check? or make it optional?

For now libsbn decides if sequences are homochronous or heterochronous by looking at the end of the taxon names. For example `A/flu/abcd_2001` is considered to be heterochronous because there is a valid "date" after the last underscore. Unfortunately some sequences can end with an underscore followed by random numbers. We need another mechanism to provide dates. It could be custom block in a nexus file, a csv file, or a switch to let know libsbn that dates are included in the taxon names

matsen:
## Description

* Allowing Newick comments next to taxon names.
* Allowing dates to be parsed from CSV via `ParseDatesFromCSV`.
* Adding separate `ParseDatesFromTaxonNames` function rather than having this happen automatically.
* Adding `initialize_time_trees` flag to these functions. This requires that the trees are in fact time trees, and initializes them for gradient calculation.
* Adding accessors to `RootedTrees` time-tree members that fail if the tree hasn't been initialized.

Closes #277 

## Tests

* A test making sure that CSV and taxon name parsing return the same results
* Tests for CSV writing and reading in general
* Updating existing tests

## Checklist:


matsen:
The following `CHECK_THROWS_AS` wasn't catching the exception for some reason, so I took it out.

```cpp
TEST_CASE("RootedSBNInstance: uninitialized time trees raise an exception") {
  auto inst = MakeFluInstance(false);
  CHECK_THROWS_AS(inst.PhyloGradients(), std::runtime_error&);
}
```

4ment:
Looks good. Can we initialize `node_bounds` even when `initialize_time_trees` is set to `false`? I use this vector to initialize the variational distribution for heterochronous datasets in phylotorch.

matsen:
Done. Please look over https://github.com/phylovi/libsbn/compare/c98ad00...277-tree-parsing and let me know what you think.

PR #287, opened 2020-10-30
278 gp optimization

junseonghwan:
## Description

This PR implements branch length optimization schedule using postorder and reverse postorder on root-source DAG. This is first of two PRs aimed at replacing leaf-source DAG by using reverse postorder traversals on root-source DAG. The next PR will be incorporating reverse postorder traversal into the two pass algorithm.

Partially addressing #278.

## Tests

- Five taxon marginal test and branch length regression tests are updated to use branch lengths estimated using the new procedure.
- Modified `GPInstance: test populate PLV` test to compare `GPEngine::GetPerGPCSPLogLikelihoods()` to its override  version `GPEngine::GetPerGPCSPLogLikelihoods(start, length)`. 

## Checklist:


src/gp_doctest.cpp
```diff
@@ -141,19 +141,6 @@ TEST_CASE("GPInstance: gradient calculation") {
   CHECK_LT(fabs(log_lik_and_derivative.second - -0.6109379521), 1e-6);
 }
 
-// Regression test.
```
matsen, line 144: It's always sad to lose a test. Is this not relevant anymore?

data/five_taxon_unrooted_with_branch_lengths.nwk
```diff
@@ -1,4 +1,4 @@
-(x0:3.48945e-06,(x2:0.825123,(x1:1.74493e-06,x3:0.825123):1.74588e-06):1.7467e-06,x4:3.00417):0;
-(x0:3.48945e-06,x1:1.74674e-06,(x3:1.74718e-06,(x2:1.74671e-06,x4:3.00417):1.7474e-06):0.825123):0;
-((x0:1.7474e-06,x1:1.74598e-06):0.8245,x2:1.7474e-06,(x3:1.74605e-06,x4:3.00417):1.74674e-06):0;
-(x3:3.49411e-06,x2:1.74532e-06,(x0:1.7474e-06,(x1:1.74463e-06,x4:3.00417):1.74532e-06):0.825123):0;
+(x0:0.825121,(x2:1.74532e-06,(x1:0.825123,x3:1.7474e-06):1.74467e-06):1.7333e-06,x4:3.00417):0;
+(x0:3.4851e-06,x1:1.7474e-06,(x3:1.74517e-06,(x2:1.7474e-06,x4:3.00417):1.74282e-06):0.825123):0;
+((x0:1.74674e-06,x1:1.7474e-06):0.820896,x2:1.74454e-06,(x3:1.7474e-06,x4:3.00417):1.74366e-06):0;
+(x3:3.49074e-06,x2:1.7474e-06,(x0:1.74464e-06,(x1:1.7474e-06,x4:3.00417):1.74113e-06):0.825123):0;
```
matsen, line 4: OK, I guess these are the new-optimized branch lengths. Note line ending.

src/gp_engine.cpp
```diff
@@ -111,12 +111,13 @@ void GPEngine::operator()(const GPOperations::UpdateSBNProbabilities& op) {
...
   }
   // else
-  EigenVectorXd log_likelihoods = GetPerGPCSPLogLikelihoods();
-  Eigen::VectorBlock<EigenVectorXd> segment =
-      log_likelihoods.segment(op.start_, range_length);
-  const double log_norm = NumericalUtils::LogSum(segment);
-  segment.array() -= log_norm;
-  q_.segment(op.start_, range_length) = segment.array().exp();
+  EigenVectorXd log_likelihood =
+      GetPerGPCSPLogLikelihoods().segment(op.start_, range_length);
```
matsen, line 115: This seems dangerous. `GetPerGPCSPLogLikelihoods` isn't an accessor-- it does calculation. Then we are returning a segment of that return value. If the full return value disappears then I think we'll have a dangling reference?

[This](https://stackoverflow.com/questions/61839333/c-dangling-reference-to-eigenmatrix-created-in-function) might be relevant.

src/gp_engine.cpp
```diff
@@ -111,12 +111,13 @@ void GPEngine::operator()(const GPOperations::UpdateSBNProbabilities& op) {
...
+  EigenVectorXd log_prior = q_.segment(op.start_, range_length).array().log();
+  EigenVectorXd unnormalize_posterior = log_likelihood + log_prior;
```
matsen, line 117: Suggest `unnormalized_posterior`.

src/gp_engine.cpp
```diff
@@ -111,12 +111,13 @@ void GPEngine::operator()(const GPOperations::UpdateSBNProbabilities& op) {
     return;
   }
   // else
-  EigenVectorXd log_likelihoods = GetPerGPCSPLogLikelihoods();
-  Eigen::VectorBlock<EigenVectorXd> segment =
-      log_likelihoods.segment(op.start_, range_length);
-  const double log_norm = NumericalUtils::LogSum(segment);
-  segment.array() -= log_norm;
-  q_.segment(op.start_, range_length) = segment.array().exp();
+  EigenVectorXd log_likelihood =
```
matsen, line 114: Suggest `log_likelihoods`.

src/gp_engine.cpp
matsen: An alternative would be to write an override of `GetPerGPCSPLogLikelihoods` which takes a start and length. Then we'd just be doing the needed calculation.

src/gp_engine.cpp
```diff
@@ -172,7 +173,7 @@ DoublePair GPEngine::LogLikelihoodAndDerivative(
   // The phylogenetic component of the likelihood is weighted with the number of times
   // we see the site patterns.
   const double log_likelihood =
-      log(q_(op.gpcsp_)) + per_pattern_log_likelihoods_.dot(site_pattern_weights_);
+      (per_pattern_log_likelihoods_.array() * site_pattern_weights_.array()).sum();
```
matsen, line 176: I'd like to know more about this change-- why we are dropping the q component. 

Also, isn't the new code equivalent to a .dot()?

matsen (changes requested):
Great! 

A few minor things. I'm looking forward to digging in more detail for https://github.com/phylovi/libsbn/issues/288 .

src/gp_dag.cpp
```diff
@@ -190,12 +193,79 @@ EigenVectorXd GPDAG::BuildUniformPrior() const {
...
+
+    if (!node->IsRoot()) {
+      UpdateRPlvs(node_id, operations);
+    }
+    if (node->IsLeaf()) {
+      continue;
+    }
+
+    if (is_reverse_postorder) {
+      OptimizeBranchLengthsUpdatePHatAndPropagateRPlv(node, true, operations);
```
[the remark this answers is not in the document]
junseonghwan: In practice, it didn't make much difference -- we could do sorted and then rotated both directions. I'll leave code comment here first and circle back to this in the other issue (#288).

src/gp_dag.cpp
```diff
@@ -291,6 +361,13 @@ void GPDAG::IterateOverRealNodes(const NodeLambda &f) const {
   }
 }
 
+void GPDAG::IterateOverLeafwardEdges(const GPDAGNode *node, bool rotated,
```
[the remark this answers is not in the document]
junseonghwan: I thought about this a bit too but decided not to -- so this version (call it version 2) takes in `NodeLambda` and the other version (version 1) that iterates over rotated and sorted takes `EdgeDestinationLambda`, call it `f`. I was thinking about adapting version 2 to call version 1 by writing NodeLambda that calls `f` but then wasn't convinced that it was cleaner.

src/gp_doctest.cpp
junseonghwan: Sort of. Regression test that involves branch length is prevalent in another test (https://github.com/phylovi/libsbn/blob/4eabe4405edcc56e82c94e0a701cd8baa30fd863/src/gp_doctest.cpp#L215-L222).

 We can certainly add it back. Let me know what you think.

src/gp_engine.cpp
junseonghwan: This is very interesting. Thanks for pointing this out. 

I agree with passing in start and length to limit the calculation only on the values that matter -- so I implemented an override for that function as you suggested.

src/gp_dag.cpp
```diff
@@ -59,7 +59,10 @@ void GPDAG::Print() const {
 
 void GPDAG::PrintGPCSPIndexer() const {
   for (const auto &[pcsp, idx] : gpcsp_indexer_) {
-    std::cout << pcsp.PCSPToString() << ", " << idx << std::endl;
+    // Since pcsp may be a rootsubsplit, we need to check and get subsplit.
```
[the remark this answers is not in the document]
junseonghwan: I've changed it to compare the bitset size to the `taxon_count_ * 3` to know if it should be `PCSPToString` or `SubsplitToString`

src/gp_dag.cpp
```diff
@@ -190,12 +193,79 @@ EigenVectorXd GPDAG::BuildUniformPrior() const {
...
+      });
+  // Update r_tilde(t) = r_hat(t) \circ p_hat(t) and r(t) = r_hat(t) \circ
+  // p_hat_tilde(t)
+  operations.push_back(Multiply{GetPLVIndex(r_plv_type, node_id),
+                                GetPLVIndex(PLVType::R_HAT, node_id),
+                                GetPLVIndex(p_plv_type, node_id)});
+}
+
+void GPDAG::ScheduleBranchLengthOptimization(SizeVector postorder_node_ids,
+                                             bool is_reverse_postorder,
```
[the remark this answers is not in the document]
junseonghwan: There will only be one DAG and postorder is root-source (leafward) and reverse postorder will just be reverse of the root-source postorder traversal.

src/gp_engine.cpp
junseonghwan: `q` is dropped because it is incorporated into `per_pattern_log_likelihoods_` in `PreparePerPatternLogLikelihoodsForGPCSP`:

https://github.com/phylovi/libsbn/blob/4eabe4405edcc56e82c94e0a701cd8baa30fd863/src/gp_engine.hpp#L170-L179

Dropping `q` here is one of the fallouts from realizing that we aren't doing exact marginal but a composite marginal. 

Yes, it is updated to use `.dot()`.

src/gp_dag.cpp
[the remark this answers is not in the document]
matsen: If we could do sorted then rotated in both directions, why have this if statement at all?

If we need to keep the logic we could do this instead:

```cpp
    OptimizeBranchLengthsUpdatePHatAndPropagateRPLV(node, is_reverse_postorder,
                                                    operations);
    OptimizeBranchLengthsUpdatePHatAndPropagateRPLV(node, !is_reverse_postorder,
                                                    operations);

```

src/eigen_sugar.hpp
```diff
@@ -47,11 +47,13 @@ void CheckVectorXdEquality(const EigenVectorXd v1, const EigenVectorXd v2,
   }
 };
 
-void CheckVectorXdEqualityAfterSorting(EigenVectorXd v1, EigenVectorXd v2,
+void CheckVectorXdEqualityAfterSorting(const EigenVectorXdRef v1, const EigenVectorXdRef v2,
```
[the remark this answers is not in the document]
matsen: Still curious.

src/gp_doctest.cpp
matsen: If there's not a compelling reason to take it out, let's leave it.

matsen (changes requested):
Thanks for the edits.

Note that I tried messing with the indices in  `GPEngine::GetPerGPCSPLogLikelihoods` like

```
log_likelihoods_.block(start + 1, 0, length, log_likelihoods_.cols())
```

which didn't make the tests fail. Could we add a test that would make this sort of perturbation fail?

src/eigen_sugar.hpp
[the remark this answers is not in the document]
junseonghwan: Yes, it was making hidden modifications. I want to keep the order unchanged so that they can be used later in the code for further testing.

src/gp_engine.cpp
matsen: ...but it doesn't look like you used it?

matsen (changes requested):
Still a problem with .segment...

PR #291, opened 2020-12-07
Set SBN parameters via a CSV of GPSCP -> double

### LINKED ISSUE #289: Serialize/deserialize SBN parameters
I'm thinking that we'll do this by being able to read in CSVs of "Pretty" SBN parameters.

We can take a vector of pretty PCSP strings as an alternate constructor to an SBNSupport.

matsen:
## Description

* Set SBN parameters via a CSV of GPSCP -> double
* Fix extra softmax bug in GP parameter reporting
* Making IndexerRepresentations more generic
* Renamed PCSPDict to PCSPCounter

Closes #289.

## Tests

Added a loading-only test and a round-trip test to and from CSV parameters.

## Checklist:


src/sbn_probability.cpp
```diff
@@ -358,14 +367,23 @@ double SBNProbability::ProbabilityOf(
   return exp(log_total_probability);
 }
 
-EigenVectorXd SBNProbability::ProbabilityOf(
+EigenVectorXd SBNProbability::ProbabilityOfCollection(
+    const EigenConstVectorXdRef sbn_parameters,
+    const std::vector<RootedIndexerRepresentation>& indexer_representations) {
+  // Lambdas aren't std:functions, so we make a std::function here.
```
matsen, line 373: @EricJIsaac @ognian- 

Do you agree that a lambda can't be used here directly? I saw https://stackoverflow.com/questions/36030589/i-cannot-pass-lambda-as-stdfunction

The `EigenVectorXdOfStdVectorT` is defined in `eigen_sugar.hpp`.

src/gp_engine.hpp
```diff
@@ -46,7 +46,7 @@ class GPEngine {
   void SetBranchLengthsToConstant(double branch_length) {
     branch_lengths_.setConstant(branch_length);
   };
-  void SetSBNParameters(EigenVectorXd q) { q_ = std::move(q); };
+  void SetSBNParameters(EigenVectorXd&& q) { q_ = std::move(q); };
```
matsen, line 49: Note @junseonghwan if we want to be able to move using this method we need to pass an rvalue reference.

matsen:
See #292 for some stuff we didn't do in this PR.

PR #298, opened 2020-12-21
Unconditional subsplit probabilities; additional outputs

### LINKED ISSUE #295: calculate probability of parent subsplits
Calculate subsplit probability using an RPO traversal as described in the "local IS" manuscript.

Proposal:

* use a DefaultDict for the values
* for each child of a PCSP, increment the weighted value for the child
* make sure that we are always keying on sorted subsplits

Then, to give context with the PCSP values, we'll want to have a version of that pretty output that shows the sorted parent.

We'll want to do this on the "true" SBN parameters.

matsen:
## Description

* Splitting SubsplitDAG out from GPDAG; now we have SubsplitDAGNode
* Calculating and outputting unconditional subsplit probabilities
* Branch lengths now available from GP using `BranchLengthsToCSV` and `branch_lengths_to_csv`

Closes #295 

## Tests

Added test with 5 taxa and 10 trees in which I can hand-verify the subsplit probabilities by counting.

## Checklist:


matsen:
Some discarded code:

```cpp
void SubsplitDAG::IterateOverRootsplits(
    const std::function<void(const Bitset &)> &f) const {
  for (const auto &rootsplit : rootsplits_) {
    f(rootsplit);
  }
}

// Iterate over the rootsplits.
void IterateOverRootsplits(const std::function<void(const Bitset &)> &f) const;
```

PR #306, opened 2021-01-19
Subsplit DAG traversal Action interface

### LINKED ISSUE #305: Refactor GP branch length optimization into an explicit DAG traversal
I think that we could have it take an object that defines 5 const lambdas, which I propose here with the code that would get moved into them:

### BeforeSubsplit(node_id)
This gets run before we perform operations on a given node:
```cpp
const auto node = GetDagNode(node_id);

  if (!node->IsRoot()) {
    UpdateRPLVs(node_id, operations);
  }

  if (node->IsLeaf()) {
    return;
  }
```

### AfterSubsplit(node_id)
This gets run after recurring down into the children of the subsplit.
```cpp
  operations.push_back(Multiply{GetPLVIndex(PLVType::P, node_id),
                                GetPLVIndex(PLVType::P_HAT, node_id),
                                GetPLVIndex(PLVType::P_HAT_TILDE, node_id)});
```

### BeforeSubsplitClade(node_id, rotated)
Before visiting a specific subsplit-clade pair, aka a subsplit with a rotation status.
```cpp
  PLVType p_hat_plv_type = rotated ? PLVType::P_HAT_TILDE : PLVType::P_HAT;
  operations.push_back(Zero{GetPLVIndex(p_hat_plv_type, node_id)});
```

### AfterSubsplitClade(node_id, rotated)
After visiting a specific subsplit-clade pair, aka a subsplit with a rotation status.
```cpp
  // Update r_tilde(t) = r_hat(t) \circ p_hat(t) and r(t) = r_hat(t) \circ
  // p_hat_tilde(t)
  operations.push_back(Multiply{GetPLVIndex(r_plv_type, node_id),
                                GetPLVIndex(PLVType::R_HAT, node_id),
                                GetPLVIndex(p_hat_plv_type, node_id)});
```

### VisitEdge(node_id, child_id, rotated)
```cpp
        OptimizeBranchLengthUpdatePHat(node_id, child_id, rotated, operations);
```

It would be more elegant to have these all wrapped into a single object.

matsen:
## Description

* Added a new DAG traversal interface with a rich Action interface, thanks to C++ wizardry from @ognian- 
* Used this to build a new GP branch length optimization method
* Removed an extra RPLV update for DAG leaves
* replacing PreOrder -> Preorder and PostOrder -> Postorder everywhere

Closes #305 

## Tests

Added operation-level test verifying that branch length traversal using the new setup is identical to that of the old.

## Checklist:

### DIFF
--- a/src/subsplit_dag.hpp
+++ b/src/subsplit_dag.hpp
@@ -73,34 +73,36 @@ class SubsplitDAG {
 
   // Traverse the DAG, applying the SubsplitDAGAction.
   template <typename SubsplitDAGActionT>
-  void IterateWithDAGAction(SubsplitDAGActionT &action) {
+  void PostOrderWithDAGAction(SubsplitDAGActionT &action) {
     std::unordered_set<size_t> visited_nodes;
     for (const auto &rootsplit : rootsplits_) {
-      IterateWithDAGActionBelowNode(action, subsplit_to_id_.at(rootsplit + ~rootsplit),
+      PostOrderWithDAGActionForNode(action, subsplit_to_id_.at(rootsplit + ~rootsplit),
                                     visited_nodes);
     }
   };
 
   // The portion of the traversal that is below a given node.
   template <typename SubsplitDAGActionT>
-  void IterateWithDAGActionBelowNode(SubsplitDAGActionT &action, size_t node_id,
+  void PostOrderWithDAGActionForNode(SubsplitDAGActionT &action, size_t node_id,
                                      std::unordered_set<size_t> visited_nodes) {
     visited_nodes.insert(node_id);
     action.BeforeNode(node_id);
-    IterateWithDAGActionBelowNodeClade(action, node_id, false, visited_nodes);
-    IterateWithDAGActionBelowNodeClade(action, node_id, true, visited_nodes);
+    PostOrderWithDAGActionForNodeClade(action, node_id, false, visited_nodes);
+    PostOrderWithDAGActionForNodeClade(action, node_id, true, visited_nodes);
     action.AfterNode(node_id);
   };
 
   // The portion of the traversal that is below a given node and a given clade.
   template <typename SubsplitDAGActionT>
-  void IterateWithDAGActionBelowNodeClade(SubsplitDAGActionT &action, size_t node_id,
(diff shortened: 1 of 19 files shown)


src/subsplit_dag.hpp
```diff
@@ -70,7 +71,58 @@ class SubsplitDAG {
...
+      DepthFirstWithActionForNode(action, subsplit_to_id_.at(rootsplit + ~rootsplit),
+                                  visited_nodes);
+    }
+  };
+
+  // The portion of the traversal that is below a given node.
+  template <typename TraversalActionT>
+  void DepthFirstWithActionForNode(const TraversalActionT &action, size_t node_id,
+                                   std::unordered_set<size_t> &visited_nodes) const {
+    action.BeforeNode(node_id);
```
junseonghwan, line 97: I just have a stupid question here:

We could call `BeforeNode` and `AfterNode` because we know that TraversalActionT passed into this function is of type `SubsplitDAGTraversalAction`, is that correct?

Is there any way to ensure that the passed type `TraversalActionT` always has these functions? I guess I'm not quite sure why generic needs to be used? Why not just use `SubsplitDAGTraversalAction`?

junseonghwan (commented):
The code looks great. I just have one naive question. Thanks!

src/subsplit_dag.hpp
matsen: We have to use a template here because `SubsplitDAGTraversalAction` isn't a class, it's a template. This is necessary because the compiler needs to understand what's going into the lambdas in order to be able to build an object.

We _could_ make it a virtual class and go that route. There are pluses and minuses.

@ognian- said in chat

> Yes, when we remove runtime typing we need to template all usage sites. The example with std::function can be used without templating, if we remove the bundled state and loose it's lifetime safety. But that's only because std::function uses virtual dispatch under the hood and this causes larger object size and dynamic allocation, leading to call overhead and cache misses. This shouldn't be allowed in the DAG traversal hot path.
> The same basic rules applies to all compile-time polymorphism, if CRTP was used it will still require templating the usage sites. C++20 has a lighter syntax for this, where a function parameter can be just declared "auto" and the compiler will turn the function into a template (trough the new concepts mechanism).

Regarding 

> Is there any way to ensure that the passed type TraversalActionT always has these functions?

Yes, it's called "concepts" and will be available to us in C++20.

PR #325, opened 2021-02-15
Replaced UnrootedPhyloGradient and RootedPhyloGradient classes with PhyloGradient

4ment:
## Description

PhyloGradient is an object containing a log_likelihood variable and a gradient variable. The gradient variable is a map containing derivatives for each sub-model.
For now the keys in the map are the variable names in the old UnrootedPhyloGradient and RootedPhyloGradient objects. Eventually the keys will be the unique identifiers of the parameters of the model.
The goal is to be add flexibility in order to add more complex models models (e.g. SRD06, phylogeography)

### DIFF
--- a/src/engine.cpp
+++ b/src/engine.cpp
@@ -75,19 +75,18 @@ std::vector<double> Engine::UnrootedLogLikelihoods(
       phylo_model_params, rescaling);
 }
 
-std::vector<UnrootedPhyloGradient> Engine::Gradients(
+std::vector<PhyloGradient> Engine::Gradients(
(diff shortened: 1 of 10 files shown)


matsen:
Thanks, @4ment! This looks just like you described on our call 👍 .

* Let's double check that we want to setting and getting things directly using a public variable. Usually I'd advocate for getter and setter methods, but here a lot of the action is going to be direct access in Python. So perhaps that means that we should be able to do the same in C++?
* It looks like your code needs to be rebased in order to merge. Do you want help with that?

PR #342, opened 2021-05-20
255 support compressed files

### LINKED ISSUE #255: support for accessing compressed files
It would be nice to be able to read from compressed files, if it wasn't too hard and didn't make our compilation process too cumbersome. 

For example, I just got this error:
```
remote: warning: File experiments/2020-07-22-frozen-2000/run.fasta.ufboot.rerooted is 63.76 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
```

I'd be happy with just reading from gzipped files, in which case I think we could just use http://zlib.net/ , right?

ognian-:
## Description

Add support for parsing gzip compressed Newick files with zlib.

Closes #255 

## Tests

Adds a test case to driver.hpp that parses the gzip-ed file at data/DS1.subsampled_10.t.nwk.gz, as well as an equivalent test for Nexus.

## Checklist:

* [x] Tests are present for new functionality, and these tests fail when perturbed


src/zlib_stream.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+namespace zlib {
+
+namespace detail {
+
+constexpr Result::Code call_zlib(int zlib_code, const z_stream& impl) {
+  if (zlib_code < 0) {
+    const auto code = static_cast<typename Exception::Code>(zlib_code);
+    if (impl.msg) {
+      throw Exception{code, impl.msg};
```
matsen, line 14: Sorry, but I think we're going to use `(` and `)` here.

src/zlib_stream.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+ZStream::~ZStream() try { close(); } catch (...) {
+}
+
+void ZStream::close() {
+  if (!closed_.test_and_set()) {
+    auto ret = detail::call_zlib(::inflateEnd(&impl_), impl_);
+    if (ret != Result::Code::ok) {
+      throw std::logic_error("Unexpected result from zlib");
```
matsen, line 43: Does this throw get silenced by the above catch?

src/zlib_stream.cpp
```diff
@@ -0,0 +1,122 @@
...
+  impl_.zalloc = nullptr;
+  impl_.zfree = nullptr;
+  impl_.opaque = this;
+  auto ret = detail::call_zlib(::inflateInit2(&impl_, 16 + MAX_WBITS), impl_);
+  if (ret != Result::Code::ok) {
+    throw std::logic_error("Unexpected result from zlib");
+  }
+}
+
+ZStream::~ZStream() try { close(); } catch (...) {
```
matsen, line 36: I'm so sorry about these nit-picks, but we use capitalized names for methods.

matsen (changes requested):
This is great, @ognian- !

I feel bad about asking for some nit-picks as part of this review. You've done an awesome job here! But I do want to keep some consistency.

Also, note that Nexus is an alternate tree format for which we'd also like to be able to parse big files. Could you add the GZ equivalent there too?

Here's a place where a test could go, which points to a file that could be used for testing as well. 
```
src/driver.hpp
110:  auto beast_nexus = driver.ParseNexusFile("data/test_beast_tree_parsing.nexus");
```

src/zlib_stream.cpp
ognian-: When Close() is called from the destructor the exception is silenced. But if it gets manually called somewhere else it should still be able to throw and allow the caller to handle the exception.

matsen (approved):

PR #352, opened 2021-08-11
Add Node Pair to DAG

### LINKED ISSUE #351: Add Node Pair to DAG
When adding a node to the DAG, we must maintain the following regarding ids and idxs:
- ids of tips are 0 through taxon_count_
- ids of parents are higher than ids of their children 
- DAG root node has the highest id
- idxs of edges descending from a given NodeClade are contiguous
- there are no gaps in node ids or edge idxs

tanviganapathy:
## Description
Add a node pair to the Subsplit DAG. Create a node_reindexer and edge_reindexer that can be used to update node ids and edge idxs.

Here is a summary of the main changes:

- Implement IsValidNodePair() which checks if it is valid to add a given pair of nodes.
- Implement AddNodePair() which adds an adjacent node pair to the DAG and builds the relevant node and edge reindexers.
- Use the above reindexers to remap all of the ids and idxs in the Subsplit DAG after adding a node pair.
- Create a Reindexer namespace with several potentially helpful Reindexer methods.
- Add/improve several Bitset methods.

Closes #351 

## Tests
- Tests for new bitset functionality.
- Tests to check AddNodePair(), UpdateNodeIds(), and UpdatePCSPIdxs().
- Tests to check AddNodePair helper methods.
- Tests to check that node_reindexer and pcsp_reindexer are correct.

## Checklist:


src/reindexer.hpp
```diff
@@ -0,0 +1,148 @@
...
+  }
+  return new_vector;
+}
+
+// Switches reverse reindexer to forward reindexer.
+// Switches forward reindexer to reverse reindexer.
+inline SizeVector SwitchReindexer(const SizeVector &reindexer) {
+  SizeVector switched_reindexer;
+  for (size_t i = 0; i < reindexer.size(); ++i) {
+    const auto switched_reindexer_i_val =
```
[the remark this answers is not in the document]
davidrich27: Looks like Tanvi already took care of it.

src/gp_doctest.cpp
```diff
@@ -613,3 +614,148 @@ TEST_CASE("GPInstance: test rootsplits") {
...
+TEST_CASE("GPInstance: IsValidNewNodePair tests") {
+  const std::string fasta_path = "data/five_taxon.fasta";
+  auto inst = GPInstanceOfFiles(fasta_path, "data/five_taxon_rooted_more_2.nwk");
+  auto& dag = inst.GetDAG();
+  // Nodes are not adjacent (12|34 and 2|4).
+  CHECK(!dag.IsValidNewNodePair(Bitset("0110000011"), Bitset("0010000001")));
+  // Nodes add/remove taxa.
+  CHECK(!dag.IsValidNewNodePair(Bitset("011000000110"), Bitset("001000000010")));
+  // Nodes already exist in the DAG (23|4 and 2|3).
+  CHECK(!dag.IsValidNewNodePair(Bitset("0011000001"), Bitset("0010000010")));
```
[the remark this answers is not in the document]
shokiami: Implemented, but no test yet.

src/reindexer.hpp
```diff
@@ -0,0 +1,148 @@
...
+// a vector of size new_size. The requirements for this are that `reindexer` is the same
+// size as new_size and contains every index in old_size exactly once.
+inline bool IsReindexer(const SizeVector &reindexer, const size_t size) {
+  if (reindexer.size() != size) {
+    return false;
+  }
+  for (size_t i = 0; i < size; ++i) {
+    const auto i_idx_itr = std::find(reindexer.begin(), reindexer.end(), i);
+    // i is in `reindexer` at least once
+    if (i_idx_itr != reindexer.end()) {
```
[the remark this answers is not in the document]
tanviganapathy: Complexity reduced to nlogn using a set.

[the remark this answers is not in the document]
shokiami: Reduced to O(n).

matsen (changes requested):

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+        CreateAndInsertEdge(parent_of_parent_id, GetDAGNodeId(parent_subsplit),
+                            rotated);
+      }
+    }
+  }
+  // Create reindexers.
+  node_reindexer = BuildNodeReindexer(prev_node_count);
+  edge_reindexer = BuildEdgeReindexer(prev_edge_count);
+  // Update new_node_ids and new_edge_idxes according to the reindexers.
+  for (size_t &id : new_node_ids) {
```
matsen, line 763: Can't we replace this with a function call?

matsen (changes requested):
Nice work! I'm going to stop partway through (time to get up 👶 x2). A few things to work on here.

Remember to update the PR description, please.

Also, paste any old code we might want to save (reverse indexer?) into a PR comment.

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+  // Only edges from an existing parent node to a new child node need to be reindexed.
+  // See SubsplitDAG::AddNodePair().
+  for (size_t edge_idx = prev_edge_count; edge_idx < GPCSPCountWithFakeSubsplits();
+       edge_idx++) {
+    size_t parent_id;
+    size_t child_id;
+    // Find the nodes that the new edge connects.
+    for (const auto &[node_pair, idx] : dag_edges_) {
+      if (idx == edge_idx) {
+        parent_id = node_pair.first;
```
matsen, line 812: check out std::tie

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+  SizeVector edge_reindexer =
+      Reindexer::IdentityReindexer(GPCSPCountWithFakeSubsplits());
+  // Only edges from an existing parent node to a new child node need to be reindexed.
+  // See SubsplitDAG::AddNodePair().
+  for (size_t edge_idx = prev_edge_count; edge_idx < GPCSPCountWithFakeSubsplits();
+       edge_idx++) {
+    size_t parent_id;
+    size_t child_id;
+    // Find the nodes that the new edge connects.
+    for (const auto &[node_pair, idx] : dag_edges_) {
```
matsen, line 810: We should use std::find_if or something here.

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+      if (idx == edge_idx) {
+        parent_id = node_pair.first;
+        child_id = node_pair.second;
+        break;
+      }
+    }
+    Bitset parent_subsplit = GetDAGNode(parent_id)->GetBitset();
```
matsen, line 817: const ref would be OK, right?

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+    Bitset child_subsplit = GetDAGNode(child_id)->GetBitset();
+    const auto idx_range = GetEdgeRange(
+        parent_subsplit, child_subsplit.SubsplitIsRotatedChildOf(parent_subsplit));
+    // New edge is added to the end of the range.
+    const size_t new_idx(idx_range.second);
+    edge_reindexer.at(edge_idx) = edge_reindexer.at(new_idx);
+    for (size_t idx = new_idx; idx < edge_idx; idx++) {
```
matsen, line 824: Shall we make this into a stand-alone function?

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+      // Every idx after the new edge's position is shifted by 1.
+      if (edge_reindexer.at(idx) >= new_idx) {
+        edge_reindexer.at(idx) = edge_reindexer.at(idx) + 1;
+      }
+    }
+  }
+  return edge_reindexer;
+}
+
+void SubsplitDAG::UpdateNodeIds(SizeVector node_reindexer) {
```
matsen, line 834: Could be const ref, right?

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+  if (node_reindexer != Reindexer::IdentityReindexer(NodeCount())) {
```
matsen, line 835: This could be simplified if we had a `NodeAdditionResult` type, right?

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+    auto dag_nodes_copy = std::move(dag_nodes_);
```
matsen, line 836: Not sure what this std::move is supposed to do. See https://www.youtube.com/watch?v=St0MNEU5b0o&ab_channel=CppCon

matsen: Oh, I see. I think you want a std::swap. Make a new one in dag_nodes_copy, then swap them.

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+    dag_nodes_ = std::vector<std::unique_ptr<SubsplitDAGNode>>(dag_nodes_copy.size());
+    for (size_t id = 0; id < NodeCount(); id++) {
+      // Update each node's fields.
+      dag_nodes_copy.at(id)->UpdateFields(node_reindexer);
+      // Update `dag_nodes_`.
+      dag_nodes_.at(node_reindexer.at(id)) = std::move(dag_nodes_copy.at(id));
```
matsen, line 842: I think this move makes sense because we are moving a unique_ptr.

src/subsplit_dag.cpp
matsen: Also, I think that a better name for this would be "ApplyNodeReindexer". What do you think?

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+    for (const auto &[node_pair, idx] : dag_edges_copy) {
+      SafeInsert(
+          dag_edges_,
+          {node_reindexer.at(node_pair.first), node_reindexer.at(node_pair.second)},
+          idx);
+    }
+  }
+}
+
+void SubsplitDAG::UpdateEdgeIdxs(SizeVector edge_reindexer) {
```
matsen, line 859: Const ref. Also, I see "idxs" here (vs "idxes") 😄

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+  if (edge_reindexer != Reindexer::IdentityReindexer(GPCSPCountWithFakeSubsplits())) {
```
matsen, line 860: As above.

src/subsplit_dag.hpp
```diff
@@ -70,14 +71,18 @@ class SubsplitDAG {
...
   std::string ToDot(bool show_index_labels = true) const;
 
   // Create a GPCSPIndexer representing the DAG.
   // The gpcsp indexer is "expanded" meaning it contains fake PCSPs and rootsplit
   // bitsets are formatted as subsplits: 1110|0001.
   BitsetSizeMap BuildGPCSPIndexer() const;
   SubsplitDAGNode *GetDAGNode(size_t node_id) const;
+  size_t GetDAGNodeId(const Bitset &subsplit) const;
   size_t DAGRootNodeId() const;
+  std::pair<size_t, size_t> GetEdgeRange(const Bitset &subsplit,
```
matsen, line 84:
```suggestion
  SizePair GetEdgeRange(const Bitset &subsplit,
```

matsen (changes requested):
Finished this phase of review! Nice work. 

I hope my comments don't seem like nit-picks, but I think that they embody principles that y'all can use for the rest of your coding career! If you think the guidelines should go some other way, let me know.

src/subsplit_dag.cpp
```diff
@@ -578,6 +608,264 @@ void SubsplitDAG::ReversePostorderIndexTraversal(
...
+    }
+  }
+  // Connect the given parent node to the given child node (don't reindex these edges if
+  // the parent is new). `prev_edge_count` is a marker conveying where we need to start
+  // reindexing edge idxes.
+  size_t prev_edge_count = GPCSPCountWithFakeSubsplits();
+  new_edge_idxes.push_back(GPCSPCountWithFakeSubsplits());
+  CreateAndInsertEdge(GetDAGNodeId(parent_subsplit), GetDAGNodeId(child_subsplit),
+                      child_subsplit.SubsplitIsRotatedChildOf(parent_subsplit));
+  if (parent_is_new) {
```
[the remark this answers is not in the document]
shokiami: Unfortunately, I don't think there is a clean way of avoiding this. I'd be happy to chat about it.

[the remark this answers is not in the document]
matsen: No problem. If we pull the body of the if statements out to methods then it should be clear enough to read.

src/subsplit_dag.cpp
shokiami: I don't think we should. We could, but it wouldn't be a reindex function call. This is simply updating the couple node ids we just added to where they are now after the reindexing. We are not reindexing new_node_ids and new_edge_idxs. I'd be happy to chat with you about this.

matsen: I understand what's going on, and that it's not an existing reindexer function call. I just suggest adding another function, which would allow us to factor out this (albeit simple) behavior and attach a useful name to the operation. This also gives an opportunity to add a test. 

And this implementation seems a little "tricky"-- we are simultaneously using id and changing it via reference. An application of std::transform would seem more canonical.

shokiami:
Here is some old "reverse reindexer" code just in case we ever need to reference it again:
```
// For "reverse reindexers", the `i`th element of the reindexer indicates the index of
// the old vector that corresponds to element `i` of the new vector. Note that this
// representation is the reverse of the forward reindexer because if index
// `i` maps to index `j`, then reindexer[`j`] = `i`.
// For example, if old_vector = [A, B, C], reindexer = [2, 0, 3, 1], and
// additional_values = [(3, D)], ApplyReverseReindexer will return [C, A, D, B].

--------

// Create a new EigenVectorXd from `old_vector` using a reverse reindexer.
// If the `i`th element of the reindexer points to an idx outside the scope of
// `old_vector`, then the new EigenVectorXd will contain the corresponding value from
// the `additional_values` map at index `i`. `additional_values` must account for
// all idxs outside the scope of `old_vector`.
// For example, if old_vector = [A, B, C], reindexer = [2, 0, 3, 1], and
// additional_values = [(3, D)], ReindexVector will return [C, A, D, B].
inline EigenVectorXd ApplyReverseReindexer(const EigenVectorXd &old_vector,
                                           const SizeVector &reverse_reindexer,
                                           const SizeDoubleMap additional_values) {
  EigenVectorXd new_vector(reverse_reindexer.size());
  for (size_t i = 0; i < reverse_reindexer.size(); i++) {
    if (reverse_reindexer[i] < old_vector.size()) {
      new_vector[i] = old_vector[reverse_reindexer[i]];
    } else {
      new_vector[i] = additional_values.at(reverse_reindexer[i]);
    }
  }
  return new_vector;
}
```

src/subsplit_dag.cpp
```diff
@@ -624,54 +623,124 @@ bool SubsplitDAG::ContainsEdge(const size_t parent_id, const size_t child_id) co
...
-  return {{true, rotated_children}, {false, sorted_children}};
+  return {rotated_children, sorted_children};
 }
 
 bool SubsplitDAG::IsValidNewNodePair(const Bitset &parent_subsplit,
                                      const Bitset &child_subsplit) const {
-  const auto &parents_of_parent = GetParentIds(parent_subsplit);
-  const auto &children_of_parent = GetChildIds(parent_subsplit);
-  const auto &children_of_child = GetChildIds(child_subsplit);
+  const auto &[rotated_parents_of_parent, sorted_parents_of_parent] =
```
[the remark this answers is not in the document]
matsen: Perhaps it's worth adding a note in the "contributing" document to make this distinction clear?

src/subsplit_dag.hpp
```diff
@@ -291,6 +294,21 @@ class SubsplitDAG {
   void RootwardDepthFirst(size_t id, SizeVector &visit_order,
                           std::unordered_set<size_t> &visited_nodes) const;
   Bitset PerhapsRotateSubsplit(const Bitset &subsplit, bool rotated) const;
+
+  // Connect the child to all of its children.
+  void ConnectChildToAllChildren(const Bitset &child_subsplit,
+                                 SizeVector &new_edge_idxs);
+  // Connect the parent to all of its children except for the given child node.
```
matsen, line 301: What is this "given child node"? The third argument is the new edge idxs.

src/subsplit_dag.cpp
```diff
@@ -802,67 +832,51 @@ SizeVector SubsplitDAG::BuildEdgeReindexer(const size_t prev_edge_count) {
...
-    }
-    // Update `parent_to_range_`.
-    for (const auto &[subsplit, idx_range] : parent_to_range_) {
-      parent_to_range_[subsplit] = {edge_reindexer[idx_range.first],
-                                    edge_reindexer[idx_range.second]};
-    }
+void SubsplitDAG::RemapEdgeIdxs(const SizeVector &edge_reindexer) {
+  // Update `dag_edges_`.
+  for (const auto &[node_pair, edge_idx] : dag_edges_) {
+    dag_edges_.at(node_pair) = edge_reindexer.at(edge_idx);
```
matsen, line 875: I wasn't really sure how/if this worked. We're modifying the thing we are looping over! It looks like it looks like it requires a little thought.

See https://stackoverflow.com/questions/8234779/how-to-remove-from-a-map-while-iterating-it ... perhaps the other links there, and either provide evidence that what we're doing is fine or make whatever modifications are necessary.

matsen: Note this happens just below as well.

src/subsplit_dag.cpp
```diff
@@ -802,67 +832,51 @@ SizeVector SubsplitDAG::BuildEdgeReindexer(const size_t prev_edge_count) {
...
-      if (idx == edge_idx) {
-        parent_id = node_pair.first;
-        child_id = node_pair.second;
-        break;
-      }
-    }
-    Bitset parent_subsplit = GetDAGNode(parent_id)->GetBitset();
-    Bitset child_subsplit = GetDAGNode(child_id)->GetBitset();
+    const auto &[node_pair, idx] =
+        *std::find_if(dag_edges_.begin(), dag_edges_.end(),
```
matsen, line 836: This is slick 😎 

But what happens if we don't find something with edge_idx here? Should we be checking for a problem there?

src/gp_doctest.cpp
```diff
@@ -613,3 +614,163 @@ TEST_CASE("GPInstance: test rootsplits") {
...
+// See diagram at https://github.com/phylovi/libsbn/issues/351#issuecomment-908711187.
+TEST_CASE("GPInstance: Only add child node tests") {
+  const std::string fasta_path = "data/five_taxon.fasta";
+  auto inst = GPInstanceOfFiles(fasta_path, "data/five_taxon_rooted_more_3.nwk");
+  auto& dag = inst.GetDAG();
+  // Before adding any nodes.
+  size_t prev_node_count = dag.NodeCount();
+  size_t prev_edge_count = dag.GPCSPCountWithFakeSubsplits();
+  // Add nodes 1|234 and 24|3 (this should just add 24|3 and associated edges).
+  const auto& [new_node_ids, new_edge_idxs, node_reindexer, edge_reindexer] =
```
matsen, line 766: Do we have tests somewhere for these return values? 😬

src/sugar.hpp
```diff
@@ -39,6 +40,7 @@ using StringDoubleVector = std::vector<std::pair<std::string, double>>;
 using DoublePair = std::pair<double, double>;
 using SizePair = std::pair<size_t, size_t>;
 using SizeOptionVector = std::vector<std::optional<size_t>>;
+using NodeAdditionResult = std::tuple<SizeVector, SizeVector, SizeVector, SizeVector>;
```
matsen, line 43: `sugar.hpp` is really for universally-useful types. Let's put this somewhere a little more local.

And I'm sorry I didn't make this clearer before, but I was hoping to make this an actual class type rather than just renaming a tuple. That will make this code a little more durable if we want to encapsulate a little functionality. However, if you don't want to do that I don't mind at all. 😄

matsen (changes requested):
Many improvements here! 

A few things that definitely need fixing, and a few more suggestions. We're getting there! 🚋

src/subsplit_dag.cpp
shokiami: Dave and I talked, and we both think that this will work because we are only modifying the values in each key-value pair and are not deleting/adding any keys.

matsen: Great! Can you provide a link describing what makes it OK?

src/sugar.hpp
shokiami: Dave and I talked, and we think that the vectors are all loosely related, and so pulling them into a single class doesn't make much sense. We did pull it out from sugar.hpp and into the class scope though!

shokiami: Actually, disregard the above. We ended up deciding to make NodeAdditionResult a class.

src/reindexer.hpp
```diff
@@ -0,0 +1,222 @@
...
+    Assert(id < reindexer.size(),
+           "The vector cannot contain an id out of bounds of the reindexer in "
+           "Reindexer::RemapIdVector.");
+  }
+  std::transform(vector.begin(), vector.end(), vector.begin(),
+                 [reindexer](size_t id) { return reindexer[id]; });
+}
+
+// In a given reindexer, take the old_id in the reindexer and reassign it to the new_id
+// and shift over all other ids to ensure that the reindexer remains valid.
```
matsen, line 96:
```suggestion
// and shift over the ids strictly between old_id and new_id to ensure that the reindexer remains valid.

```

matsen: (I guess this comment is going to have to get reformatted too.)

matsen (approved):
YAY! I'm going to approve this for merging because it's completed its scope of work. (I do have one little suggestion below).

Congratulations!

PR #357, opened 2021-08-20
281 beagle thread exceptions

### LINKED ISSUE #281: Exceptions from threads aren't propagated to the main thread.

davidrich27:
## Description

This change resolves the issue of exceptions occurring during TaskProcessor's multithreading not properly handling when exceptions occurred on child threads.  To fix this, an exception pointer was added to TaskProcessor.  Now, when an exception occurs, further work is ceased by worker threads.  After threads rejoin main thread in Wait(), then CheckExceptions() checks whether an exception occurred on any child.  Only first exception is captured and reported.  This is thrown by TaskProcessor to be caught and handled by FatBeagleParallelize().  

Closes #281 

## Tests

This functionality is tested by TEST_CASE("RootedSBNInstance: uninitialized time trees raise an exception"), at rooted_sbn_instance.hpp:405.  

## Checklist:


src/fat_beagle.hpp
```diff
@@ -137,14 +138,19 @@ std::vector<TOut> FatBeagleParallelize(
...
-  TaskProcessor<FatBeagle *, size_t> task_processor(
-      std::move(fat_beagle_queue), std::move(tree_number_queue),
-      [&results, &tree_collection, &param_matrix, &rescaling, &f](FatBeagle *fat_beagle,
-                                                                  size_t tree_number) {
-        fat_beagle->SetParameters(param_matrix.row(tree_number));
-        fat_beagle->SetRescaling(rescaling);
-        results[tree_number] = f(fat_beagle, tree_collection.GetTree(tree_number));
-      });
+
+  std::unique_ptr<TaskProcessor<FatBeagle *, size_t>> task_processor =
```
matsen, line 142: would `auto` work here?

src/fat_beagle.hpp
```diff
@@ -137,14 +138,19 @@ std::vector<TOut> FatBeagleParallelize(
...
+    std::make_unique<TaskProcessor<FatBeagle *, size_t>>(
+    std::move(fat_beagle_queue), std::move(tree_number_queue),
+    [&results, &tree_collection, &param_matrix, &rescaling, &f](FatBeagle *fat_beagle,
+                                                                size_t tree_number) {
+      fat_beagle->SetParameters(param_matrix.row(tree_number));
+      fat_beagle->SetRescaling(rescaling);
+      results[tree_number] = f(fat_beagle, tree_collection.GetTree(tree_number));
+    });
+    
+  task_processor->Wait();
```
matsen, line 152: Note that we need to update the docs at the top of this file now that we need to explicitly Wait.

src/task_processor.hpp
```diff
@@ -53,9 +53,13 @@ class TaskProcessor {
         task_(task),
         threads_(executor_queue.size()) {
     // Make as many threads as there are executors.
+    threads_rejoined_ = false;
```
matsen, line 56: This line seems a little out of place relative to the comment above.

src/task_processor.hpp
```diff
@@ -101,7 +124,20 @@ class TaskProcessor {
         // Unlock now that we're done messing with the queues.
         lock.unlock();
         // Run task.
-        task_(executor, work);
+        try {
+          task_(executor, work);
+        } 
+        catch (const std::exception &exception) {
+          // If the task throws an exception, record it for the master thread
```
matsen, line 131:
```suggestion
          // If the task throws an exception, record it for the master thread.
```

src/task_processor.hpp
```diff
@@ -65,7 +69,10 @@ class TaskProcessor {
   TaskProcessor &operator=(const TaskProcessor &&) = delete;
 
   // Ensure that all tasks are done before destructor returns.
-  ~TaskProcessor() { Wait(); }
+  ~TaskProcessor() { 
+    // If threads were not rejoined earlier, do it now.
+    if ( threads_rejoined_  == false) { Wait(); }
```
matsen, line 74: Did you clang format this? 🧐

src/task_processor.hpp
```diff
@@ -53,9 +53,13 @@ class TaskProcessor {
         task_(task),
         threads_(executor_queue.size()) {
     // Make as many threads as there are executors.
+    threads_rejoined_ = false;
     for (size_t i = 0; i < threads_.size(); i++) {
       threads_[i] = std::thread(&TaskProcessor::thread_handler, this);
     }
+    // Exception pointer is null unless exception occurs on one or more executors/threads.
+    exception_ptr_ = nullptr;
```
matsen, line 61: What do you think about moving these assignments to where these variables were declared?

src/task_processor.hpp
```diff
@@ -53,9 +53,13 @@ class TaskProcessor {
...
+    exception_occurred_ = false;
```
matsen, line 62: I wonder if this variable is necessary? Could we just test if the exception_ptr_ is nullptr?

src/task_processor.hpp
```diff
@@ -75,6 +82,16 @@ class TaskProcessor {
         threads_[i].join();
       }
     }
+    threads_rejoined_ = true;
+    // When all thread work completed and rejoined, check that no child threads had exceptions.
```
matsen, line 86: I'm going to suggest that this comment and the one in CheckExceptions could be eliminated if we renamed CheckExceptions to something like `RethrowAnyExceptionsEncounteredInChildThread`.

Or, perhaps this method could be eliminated? The code is very clear.

matsen (changes requested):
This looks good!

Re comments, this does a reasonable summary of my stance: https://www.arhohuttunen.com/code-comments/

src/fat_beagle.hpp
davidrich27: Should have thought of it sooner.  There's no reason we can't move Wait() to the end of the Constructor rather than the Destructor is there?  It still passes the test case with this reorganization.

matsen: Interesting idea! And this doesn't have the problem you found before because at the end of the constructor we aren't going out of scope?

src/task_processor.hpp
davidrich27: Whoops, missed a file.  So I can just run `make format` for this right?

src/fat_beagle.hpp
davidrich27: Correct.

src/task_processor.hpp
matsen: yep

src/fat_beagle.hpp
matsen: Love it

src/task_processor.hpp
davidrich27: Agreed. Done.

src/task_processor.hpp
davidrich27: Agreed. Also done.

src/task_processor.hpp
davidrich27: I agree it is clear.  I just merged it into the end of Wait() and eliminated the method.

src/task_processor.hpp
davidrich27: Fixed.

src/task_processor.hpp
```diff
@@ -84,13 +89,19 @@ class TaskProcessor {
...
+
+  std::exception_ptr exception_ptr_ = nullptr;
+  std::mutex exception_lock_;
 
   void thread_handler() {
     std::unique_lock<std::mutex> lock(lock_);
-    while (work_queue_.size()) {
+    // Continue allocating free executors for work until no more work or executor
+    // exception occurs.
+    while (work_queue_.size() && exception_ptr_ == nullptr) {
```
ognian-, line 101: exception_ptr_ is accessed here without mutex guard

src/task_processor.hpp
davidrich27: This variable was optimized out.

src/fat_beagle.hpp
davidrich27: Yes

PR #365, opened 2021-09-08
Revert "Add Node Pair to DAG"

davidrich27:
Reverts phylovi/libsbn#352.


davidrich27:
Bug created while merging with main branch due to changes in function definition.  Bug was resolved and new commit squashed, but branch needs to be reverted and recommitted.

davidrich27:
Issue with #352 resolved by Erick Matsen.

PR #388, opened 2021-12-07
Topology sampling on SubsplitDAG

ognian-:
## Description

A new class called TopologySampler is added to sample trees from SubsplitDAG, and *SBNInstance as well.

## Checklist:


src/topology_sampler.hpp
```diff
@@ -0,0 +1,129 @@
...
+#pragma once
+
+#include "node.hpp"
+#include "mersenne_twister.hpp"
+#include "eigen_sugar.hpp"
+
+class TopologySampler {
+public:
+
+  class Input {
```
matsen, line 13: This is very cool! @ognian- -- is there a name for this pattern?

src/topology_sampler.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+private:
+  const SubsplitDAG &dag_;
+  EigenConstVectorXdRef sbn_parameters_;
+};
+
+#ifdef DOCTEST_LIBRARY_INCLUDED
+
+#include "doctest_constants.hpp"
+#include "unrooted_sbn_instance.hpp"
```
matsen, line 84: Wow!!

If I understand correctly, this doesn't set up a circular dependency because this code isn't part of `topology_sampler.hpp` and really just gets pulled into the doctest. Is that right?

ognian-: There isn't a circular dependency, but it's written like this to avoid bringing in unnecessary symbols during normal usage (outside doctest). If there was a circular dependency, it would be triggered during doctest compilation.

src/topology_sampler.hpp
ognian-: I'm not aware of a pattern name for this particular setup, it's simply an interface class for the algorithm input. I chose run-time polymorphism instead of template because the overhead is minimal, but it will reduce compilation time and header dependencies.

matsen:
@ognian- I am sorry to make you revert some of these changes, but I talked with the rest of the crew and we'd like for all of the SBN parameters to be held in log space. So, could you please take out the flags that you added for that? 

Then, in the test, can you just get the prior vector and then apply log to it, and pass that?

src/topology_sampler.cpp
```diff
@@ -0,0 +1,80 @@
...
+  return dag_.GetDAGNode(dag_.RootsplitIds().at(rootsplit_idx))->GetBitset();
+  ;
+}
+
+const SizePair &SubsplitDAGSamplerInput::ParentToRangeAt(const Bitset &parent) const {
+  return dag_.ParentToRange().at(parent);
+}
+
+const Bitset &SubsplitDAGSamplerInput::IndexToChildAt(size_t child_idx) const {
+  for (auto &&i : dag_.DAGEdges()) {
```
matsen, line 76: Should this `&&` become standard or discouraged? 

https://stackoverflow.com/questions/13130708/what-is-the-advantage-of-using-forwarding-references-in-range-based-for-loops/13130795#13130795

src/topology_sampler.cpp
```diff
@@ -0,0 +1,80 @@
...
+const Bitset &SubsplitDAGSamplerInput::RootsplitsAt(size_t rootsplit_idx) const {
+  return dag_.GetDAGNode(dag_.RootsplitIds().at(rootsplit_idx))->GetBitset();
+  ;
+}
+
+const SizePair &SubsplitDAGSamplerInput::ParentToRangeAt(const Bitset &parent) const {
+  return dag_.ParentToRange().at(parent);
+}
+
+const Bitset &SubsplitDAGSamplerInput::IndexToChildAt(size_t child_idx) const {
```
matsen, line 75: This is sort of an interesting case where it's not an "At"... this is a linear time lookup.

src/topology_sampler.cpp
```diff
@@ -0,0 +1,80 @@
...
+// The input to this function is a parent subsplit (of length 2n).
+Node::NodePtr TopologySampler::SampleTopology(const Input &input,
+                                              const Bitset &parent_subsplit) const {
+  auto process_subsplit = [this, &input](const Bitset &parent) {
+    auto singleton_option = parent.SubsplitGetClade(1).SingletonOption();
+    if (singleton_option) {
+      return Node::Leaf(*singleton_option);
+    }
+    auto child_index = SampleIndex(input, input.ParentToRangeAt(parent));
+    return SampleTopology(input, input.IndexToChildAt(child_index));
```
matsen, line 54: Using this linear time function means that we the whole topology sampling is quadratic time when it should be linear. 

Let's think a little more about exactly how we want to set up this topology sampling-- I think that we might need a subsplit DAG-specific method rather than doing something generic here. 

The range of indices that we want should be available as the leafward edges from the node, right? See the documentation in `subsplit_dag.hpp`: 

```
// - The nodes of the DAG have vectors of indices representing their edges to other
// nodes. These include edges to the children (leafward edges) and also the edges that
// are connecting to that node (rootward edges). Furthermore, these sets of edges are
// separated into two classes: those that split apart the left component of the subsplit
// ("rotated" edges) and those that split apart the right component of the subsplit
// ("sorted" edges). The nodes of the DAG also include bitsets describing the subsplit
// that they represent.
```

matsen (commented):
Have a look at the say subsplit DAGs are laid out and we can talk on Monday!

src/subsplit_dag.cpp
```diff
@@ -322,8 +326,8 @@ std::vector<Bitset> SubsplitDAG::GetSortedVectorOfNodeBitsets() const {
 std::vector<Bitset> SubsplitDAG::GetSortedVectorOfEdgeBitsets() const {
   std::vector<Bitset> edges;
   for (auto i : storage_.GetLines()) {
-    auto parent_bitset = GetDAGNode(i.GetParent()).GetBitset();
-    auto child_bitset = GetDAGNode(i.GetChild()).GetBitset();
+    auto parent_bitset = i.GetParentNode().GetBitset();
```
matsen, line 329: I let this slide on the first subsplit DAG PR refactor, but I really do prefer more informative variable names everywhere, including in situations like this.

src/subsplit_dag_node.hpp
```diff
@@ -31,18 +31,23 @@ static inline void RemapNeighbors(NeighborsView neighbors,
                                   const SizeVector& node_reindexer) {
   std::map<VertexId, LineId> remapped;
   for (auto i = neighbors.begin(); i != neighbors.end(); ++i) {
-    remapped[node_reindexer[*i]] = i.GetEdge();
+    remapped[node_reindexer[*i]] = i.GetEdgeId();
```
matsen, line 34: I do prefer `.at()` rather than brackets for access.

src/subsplit_dag_node.hpp
```diff
@@ -31,18 +31,23 @@ static inline void RemapNeighbors(NeighborsView neighbors,
...
   }
   neighbors.SetNeighbors(remapped);
 }
 }  // namespace
 
 template <typename T>
 class GenericSubsplitDAGNode {
+  constexpr static bool is_const = std::is_const_v<T>;
+  using dag_storage_type =
```
matsen, line 43: Types are CamelCase

src/subsplit_dag_storage.hpp
```diff
@@ -109,13 +109,34 @@ class DAGLineStorage : public DAGLine<DAGLineStorage> {
...
+class DAGVertex;
+template <typename>
+class GenericSubsplitDAGNode;
+using MutableSubsplitDAGNode = GenericSubsplitDAGNode<DAGVertex>;
+using SubsplitDAGNode = GenericSubsplitDAGNode<const DAGVertex>;
+
 template <typename T>
 class DAGLineView : public DAGLine<DAGLineView<T>> {
+  constexpr static bool is_const = std::is_const_v<T>;
+  using storage_type =
```
matsen, line 122: CamelCase as above

src/subsplit_dag_storage.hpp
```diff
@@ -161,40 +183,52 @@ struct tuple_element<1, ::ConstLineView> {
...
 class GenericNeighborsView {
+  constexpr static bool is_const = std::is_const_v<T>;
   using iterator_type =
-      std::conditional_t<std::is_const_v<T>, std::map<VertexId, LineId>::const_iterator,
+      std::conditional_t<is_const, std::map<VertexId, LineId>::const_iterator,
                          std::map<VertexId, LineId>::iterator>;
-  using map_type =
-      std::conditional_t<std::is_const_v<T>, const std::map<VertexId, LineId>,
-                         std::map<VertexId, LineId>>;
+  using map_type = std::conditional_t<is_const, const std::map<VertexId, LineId>,
```
matsen, line 190: Here too

src/subsplit_dag_storage.hpp
```diff
@@ -376,14 +411,14 @@ class SubsplitDAGStorage {
     if (parent >= vertices_.size() || child >= vertices_.size()) return {};
     auto result = vertices_.at(parent).FindNeighbor(child);
     if (!result.has_value()) return {};
-    return lines_.at(std::get<0>(result.value()));
+    return ConstLineView{lines_.at(std::get<0>(result.value())), *this};
   }
 
   std::optional<ConstLineView> GetLine(LineId id) const {
     if (id >= lines_.size()) return {};
     auto& line = lines_[id];
```
matsen, line 419: Do we want [] access vs .at() here?

src/topology_sampler.hpp
```diff
@@ -0,0 +1,164 @@
+// Copyright 2019-2021 bito project contributors.
```
matsen, line 1:
```suggestion
// Copyright 2019-2022 bito project contributors.
```

matsen: Please search for other instances.

matsen (approved):
Awesome! This is looking really nice.

I'm approving this because I don't want to block merging, but I'd appreciate a few little tweaks here.

PR #400, opened 2022-01-24
242 gp hotstart rootedpcpspreorder

### LINKED ISSUE #376: gather all branch lengths from a sample of trees
For our GP branch length comparison, we are going to want to compare branch lengths inferred by GP to those from an external collection of trees (e.g. a MB run). We have good infrastructure for getting the per-PCSP mean of branch lengths for a collection of trees using the "hotstart" code.

* At very least we are going to need to do something to get those mean branch lengths. This could be done by setting hot start branch lengths then exporting. 
* If we want the distribution of such branch lengths we definitely need some additional code.

hrnasif:
## Description

Modify `Node::RootedPCSPPreorder` with the flag `allow_leaves`, that gives the option to traverse to the leaves. No default behavior set. Coded as false whenever used, except for `hotstart`-like functions.

Implemented `GPEngine::FunctionOverRootedTreeCollection` that generalizes traversing inputted RootedTrees and calculating some quantity. Modified `GPEngine::HotStartBranchLengths` and added `GPEngine::GatherBranchLengths` based on the general method.

Modified hotstart test in `gp_doctest` to now test HotStart and Gather for both internal and pendant branches.

Closes #376

## Tests

Please describe the tests added to verify correct behavior.
Please double check that you can make your tests fail by perturbing them.

## Checklist:


src/gp_engine.hpp
```diff
@@ -79,6 +79,9 @@ class GPEngine {
   // Use branch lengths from loaded sample as a starting point for optimization.
   void HotStartBranchLengths(const RootedTreeCollection& tree_collection,
                              const BitsetSizeMap& indexer);
+  // Gather branch lengths from loaded sample with their corresponding pcsp.
+  SizeDoubleVector GatherBranchLengths(const RootedTreeCollection& tree_collection,
```
matsen, line 83: Rather than returning a SizeDoubleVector of all of the branch lengths jumbled up, it would seem nicer to return a map from size (PCSP) to a double vector, collating all of the branch lengths together for a given PCSP. Is there a reason we shouldn't do that?

src/sbn_maps.cpp
```diff
@@ -67,15 +67,20 @@ Bitset SBNMaps::PCSPBitsetOf(const size_t leaf_count,  //
...
   bitset.CopyFrom(sister_node->Leaves(), 0, sister_direction);
   bitset.CopyFrom(focal_node->Leaves(), leaf_count, focal_direction);
-  auto child0_bitset = child0_node->Leaves();
-  if (child0_direction) {
-    child0_bitset.flip();
-  }
-  auto child1_bitset = child1_node->Leaves();
-  if (child1_direction) {
-    child1_bitset.flip();
+  if (child0_node == nullptr || child1_node == nullptr) {
```
matsen, line 70: Let's update the .hpp documentation to describe this behavior.

matsen (commented):
Looks good, @hrnasif ! Just one question and a comment here.

matsen:
@junseonghwan here is a little script that uses this in its Python form. Note the `branch_lengths = {k: np.array(v) for k, v in inst.gather_branch_lengths().items()}`, which means that the dictionary contains numpy arrays. Hopefully that works for you.

```
import bito
import numpy as np

tree_path = "data/hotstart_bootstrap_sample.nwk"
inst = bito.gp_instance("_ignore/mmapped_plv.data")
inst.read_fasta_file("data/hotstart.fasta")
inst.read_newick_file(tree_path)
inst.make_engine()

branch_lengths = {k: np.array(v) for k, v in inst.gather_branch_lengths().items()}
branch_lengths[13]
```

matsen (approved):

PR #407, opened 2022-02-22
403 sdag storage

### LINKED ISSUE #403: Store edge ids with neighboring nodes
In SubsplitDAGNode add edge ids to the lists of neighbor node ids.

ognian-:
## Description

Refactor the internal edge and node storage of SubsplitDAG to allow for more flexible navigation of it's components.

Closes #403 

## Tests

"SubsplitDAGStorage: LinesView structured binding" - tests using the edges view in a ranged for-loop with structured bindings for pair<nodeId, nodeId> and an edgeId.

"SubsplitDAGStorage: Neighbors iterator" - read and write functionality of the custom node neighbors iterator

## Checklist:


src/subsplit_dag_node.hpp
```diff
@@ -18,112 +18,231 @@
...
+  size_t parent_node_count = GetRightRootward().size() + GetLeftRootward().size();
+  if (parent_node_count > 0) {
+    if (GetRightLeafward().size() == 0 || GetRightLeafward().size() == 0) {
+      return false;
+    }
+  } else {
+    if (GetRightLeafward().size() > 0 || GetRightLeafward().size() > 0) {
+      return false;
+    }
+  }
```
davidrich27, line 144: Error on my end:  
```suggestion
    if (GetRightLeafward().size() == 0 || GetLeftLeafward().size() == 0) {
      return false;
    }
  } else {
    if (GetRightLeafward().size() > 0 || GetLeftLeafward().size() > 0) {
      return false;
    }
  }
```

src/subsplit_dag.cpp
```diff
@@ -1179,33 +1195,37 @@ SizeVector SubsplitDAG::BuildEdgeReindexer(const size_t prev_edge_count) {
...
 void SubsplitDAG::RemapNodeIds(const SizeVector &node_reindexer) {
+  std::vector<DAGVertex> nodes = {storage_.GetVertices().begin(), storage_.GetVertices().end()};
+  std::vector<DAGVertex> nodes_copy = Reindexer::Reindex(nodes, node_reindexer);
+  storage_.SetVertices(nodes_copy);
+
   // Update `dag_nodes_`.
-  auto dag_nodes_copy = Reindexer::Reindex(dag_nodes_, node_reindexer);
-  dag_nodes_.swap(dag_nodes_copy);
+  // auto dag_nodes_copy = Reindexer::Reindex(dag_nodes_, node_reindexer);
+  // dag_nodes_.swap(dag_nodes_copy);
```
davidrich27, line 1204: Need to remove?

src/subsplit_dag.hpp
```diff
@@ -129,13 +129,14 @@ class SubsplitDAG {
   // Get Taxon's bitset clade positional id.
   size_t GetTaxonId(const std::string &name) const;
   // Get node based on node id.
-  SubsplitDAGNode *GetDAGNode(const size_t node_id) const;
+  SubsplitDAGNode GetDAGNode(const size_t node_id) const;
+  MutableSubsplitDAGNode GetDAGNode(const size_t node_id);
```
davidrich27, line 133: For my clarification:
So it will use the always use the mutable `GetDAGNode()` if SubsplitDAG is not `const`, is that correct?

src/subsplit_dag.cpp
```diff
@@ -124,14 +125,14 @@ StringSizeMap SubsplitDAG::SummaryStatistics() const {
...
 }
 
 void SubsplitDAG::PrintNodes() const {
-  for (const auto &dag_node : dag_nodes_) {
-    std::cout << dag_node->Id() << ": " << dag_node->GetBitset().SubsplitToString()
+  for (const auto &dag_node : storage_.GetVertices()) {
+    std::cout << dag_node.Id() << ": " << dag_node.GetBitset().SubsplitToString()
               << std::endl;
   }
 }
```
davidrich27, line 138: Is there a reason `dag_node`'s `const` ref was dropped in one function but not the other?

src/subsplit_dag.cpp
```diff
@@ -338,10 +346,10 @@ EigenVectorXd SubsplitDAG::BuildUniformOnTopologicalSupportPrior() const {
 Node::NodePtrVec SubsplitDAG::GenerateAllTopologies() const {
   std::vector<Node::NodePtrVec> topology_below(NodeCount());
 
-  auto GetSubtopologies = [&topology_below](const SubsplitDAGNode *node) {
+  auto GetSubtopologies = [&topology_below](SubsplitDAGNode node) {
     Node::NodePtrVec rotated_subtopologies, sorted_subtopologies;
     for (const bool rotated : {false, true}) {
-      for (const auto &child_id : node->GetLeafward(rotated)) {
+      for (const auto &child_id : node.GetLeafward(rotated)) {
```
davidrich27, line 352: We are dropping all `const` refs for iterating over `NeighborView`s as well?

src/subsplit_dag.cpp
```diff
@@ -400,19 +408,19 @@ Node::NodePtrVec SubsplitDAG::GenerateAllTopologies() const {
 
 EigenVectorXd SubsplitDAG::BuildUniformOnAllTopologiesPrior() const {
   EigenVectorXd result = EigenVectorXd::Zero(EdgeCountWithLeafSubsplits());
-  for (const auto &[parent_child_id, edge_idx] : dag_edges_) {
+  for (const auto &[parent_child_id, edge_idx] : storage_.GetLines()) {
```
davidrich27, line 411: I presume also should drop `const` from `LineView`s?

davidrich27:
The API look pretty good to me.  I got build errors on the branch, even will errors off.  It looks like doctest still has pointer notation for accessing `SubsplitDAGNode`s.  And I saw some other const refs when iterating over views, but I will wait for a response from Ognian to see if there is some reason to keep them in before flagging.

src/subsplit_dag.hpp
```diff
@@ -454,7 +455,7 @@ class SubsplitDAG {
   size_t CreateAndInsertEdge(const size_t parent_id, const size_t child_id,
                              bool rotated);
   // Add edge between given parent and child nodes to the DAG.
-  void ConnectGivenNodes(const size_t parent_id, const size_t child_id, bool rotated);
+  void ConnectGivenNodes(const size_t parent_id, const size_t child_id, bool rotated, LineId edge_id);
```
matsen, line 458: I wonder if the `Line` terminology could be limited to just the internals of the storage. Can we just use `EdgeId` here?

src/subsplit_dag_node.hpp
```diff
@@ -18,112 +18,231 @@
...
-    AddEdge(adjacent_node_id, is_leafward, is_rotated);
-  }
-  void AddLeftLeafward(size_t node_id) { left_leafward_.push_back(node_id); }
-  void AddRightLeafward(size_t node_id) { right_leafward_.push_back(node_id); }
-  void AddLeftRootward(size_t node_id) { left_rootward_.push_back(node_id); }
-  void AddRightRootward(size_t node_id) { right_rootward_.push_back(node_id); }
 
   // Get vector of all adjacent node vectors along the specified direction.
-  const SizeVector &GetEdge(bool is_leafward, bool is_rotated) {
+  ConstNeighborsView GetEdge(bool is_leafward, bool is_rotated) const {
```
matsen, line 56: Interesting that this returns a NeighborsView. I see how this is a nice consistent choice, but does it make it difficult to consume?

matsen (commented):
I didn't have as much time as I wanted to look through this morning, but will continue to do so and also we get to meet today.

But this branch doesn't compile for me... lots of errors in doctest.

src/subsplit_dag.cpp
ognian-: Yes, I forgot those

src/subsplit_dag.hpp
ognian-: Correct. I chose to keep the name SubsplitDAGNode instead of ConstSubsplitDAGNode to avoid changing too many locations. In the future I'd like to go trough the codebase and ensure const-correctness of all SubsplitDAGNode usage locations.

src/subsplit_dag.cpp
ognian-: I'd like to drop it everywhere, because SubsplitDAGNode now is a lightweight object, meant to be passed by value like an iterator or string_view.

src/subsplit_dag.cpp
ognian-: Yes, GetLeafward() returns [Const]NeighborsView, which is similar to a string_view: can be cheaply passed by value.

src/subsplit_dag.hpp
ognian-: You're correct, I'll change it

src/subsplit_dag_node.hpp
ognian-: [Const]NeighborsView looks to the users similarly to a SizeVector - can be iterated, and queried for size() and empty(). It can easily be extended to support indexed access with [i] and at(i) like LinesView. It wasn't needed so far by the rest of the code so I didn't add those, but I'd like to add them in the future.

src/subsplit_dag.cpp
```diff
@@ -691,38 +701,52 @@ void SubsplitDAG::BuildDAGEdgesFromEdgeIndexer(BitsetSizeMap &edge_indexer) {
            "All edges should be bitsets with size 3 times taxon_count_.");
     const auto parent_id = GetDAGNodeId(edge.PCSPGetParentSubsplit());
     const auto child_id = GetDAGNodeId(edge.PCSPGetChildSubsplit());
-    SafeInsert(dag_edges_, {parent_id, child_id}, index);
+    auto left = GetDAGNode(parent_id).GetLeftLeafward();
+    auto clade = std::find(left.begin(), left.end(), child_id) != left.end()
```
matsen, line 705: Should we assert that it appears in at least one of the clades? And would it be worth factoring this out into a function?

src/subsplit_dag.cpp
```diff
@@ -691,38 +701,52 @@ void SubsplitDAG::BuildDAGEdgesFromEdgeIndexer(BitsetSizeMap &edge_indexer) {
...
-                     EdgeCountWithLeafSubsplits());
+          storage_.AddLine({EdgeCountWithLeafSubsplits(), node.Id(),
+                            GetDAGNodeId(current_bitset),
+                            rotated ? Clade::Left : Clade::Right});
         });
   }
 }
 
+void SubsplitDAG::StoreEdgeIds() {
+  for (auto i : storage_.GetLines()) {
```
matsen, line 728: I would suggest a more descriptive variable name here. And we don't want a const ref here?

src/subsplit_dag.hpp
```diff
@@ -129,13 +129,14 @@ class SubsplitDAG {
...
-  SubsplitDAGNode *GetDAGNode(const size_t node_id) const;
+  SubsplitDAGNode GetDAGNode(const size_t node_id) const;
+  MutableSubsplitDAGNode GetDAGNode(const size_t node_id);
   // Get the node id based on the subsplit bitset.
   size_t GetDAGNodeId(const Bitset &subsplit) const;
   // Gets the node id of the DAG root.
   size_t DAGRootNodeId() const;
   // Return the node ids corresponding to the rootsplits.
-  const SizeVector &RootsplitIds() const;
+  ConstNeighborsView RootsplitIds() const;
```
matsen, line 139: This appears in the public interface but isn't documented anywhere that I can find. What is it, and how can I use it? Just a few lines in this header file should be fine.

src/subsplit_dag.hpp
```diff
@@ -487,21 +489,17 @@ class SubsplitDAG {
   // Expand dag_edges_ and parent_to_child_range_ with leaf subsplits at the end.
   void AddLeafSubsplitsToDAGEdgesAndParentToRange();
 
+  void StoreEdgeIds();
```
matsen, line 492: Is this really still here?

src/subsplit_dag_node.hpp
matsen: This looks like just what is needed for the documentation suggested above.

src/subsplit_dag.cpp
matsen: Still?

matsen (changes requested):
Pass through header files.

src/subsplit_dag.hpp
ognian-: I was thinking about GetStorage(), which I moved to an anonymous namespace. I'll fix this too

src/subsplit_dag.cpp
ognian-: I'm adding the assert. We already have a function that does that: FindNeighbor(). It returns `std::optional<std::tuple<LineId, Direction, Clade>>` which contains every possible information about the lookup: if it's found or not, it's edge id, direction and clade. We can't use FindNeighbor() here, because edges hasn't been added yet.

src/subsplit_dag.hpp
ognian-: I've added the doc comment on SubsplitDAGNode::GetEdge, maybe I should move it to the declaration of NeighborsView and add reference comments to all functions that return it?

matsen: That sounds good. You don't need to refer to the docs in every function that returns it... just something at the top of the header file for subsplit DAG and subsplit DAG node would do. thanks! ✨

ognian-: I've moved the usage information to SubsplitDAGNode top section

davidrich27:
Got this warning during build:
```
/home/drich/matsen-lab/bito/src/subsplit_dag_node.hpp:235:31: error: unused variable ‘line_id’ [-Werror=unused-variable]
   for (auto [node_ids, line_id] : storage.GetLines()) {
```

matsen (approved):

PR #409, opened 2022-02-25
Enable loading of duplicates of the first loaded tree.

### LINKED ISSUE #408: Function for duplicating trees
It would be nice to be able to duplicate tree(s) loaded in BITO.

For example, since torchtree makes BITO load a single topology it cannot do multithreading. With a multisample ELBO torchtree currently calculates the K gradients one after another.

We could have a `DuplicateTree(size K)` function in `GenericSBNInstance`. It has a similar function for model parameters: `ResizePhyloModelParams`

matsen:
## Description

Enable loading of duplicates of the first loaded tree.

Closes #408 

## Tests

Tests that trees are properly duplicated, and that an exception is raised when there are no trees.

## Checklist:

* [ x] Code follows our detailed [contribution guidelines](CONTRIBUTING.md)
* [ x] `clang-format` has been run
* [ x] TODOs have been eliminated from the code
* [ x] Comments are up to date, document intent, and there are no commented-out code blocks

### DIFF
--- a/src/generic_sbn_instance.hpp
+++ b/src/generic_sbn_instance.hpp
@@ -279,6 +279,11 @@ class GenericSBNInstance {
   void SetAlignment(const Alignment &alignment) { alignment_ = alignment; }
   void SetAlignment(Alignment &&alignment) { alignment_ = alignment; }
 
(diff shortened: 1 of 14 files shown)


4ment (approved):
Looks great!

PR #418, opened 2022-03-16
417 graft sdag storage

### LINKED ISSUE #417: Graft support in SubsplitDAGStorage
Add support for graft operations in SubsplitDAGStorage

ognian-:
## Description

Move the graft node and edge accessing logic to SubsplitDAGStorage, and make GraftDAG inherit from SubsplitDAG. This replaces the previous fragile implementation with Id offsets, and reduces duplication of methods between GraftDAG and SubsplitDAG. 

Closes #417 

## Tests

Added additional testing to TEST_CASE("NNI Engine: GraftDAG").

## Checklist:

### DIFF
--- a/src/subsplit_dag.cpp
+++ b/src/subsplit_dag.cpp
@@ -38,6 +38,27 @@ SubsplitDAG::SubsplitDAG(size_t taxon_count,
          "Taxon ids do not cover all indices from 0 to taxon_count.");
 }
 
+SubsplitDAG::SubsplitDAG(SubsplitDAG &host_dag, HostDispatchTag)
+    : storage_{host_dag.storage_, HostDispatchTag()},
+      dag_taxa_{host_dag.dag_taxa_},
+      subsplit_to_id_{host_dag.subsplit_to_id_},
+      parent_to_child_range_{host_dag.parent_to_child_range_},
+      taxon_count_{host_dag.taxon_count_},
+      edge_count_without_leaf_subsplits_{host_dag.edge_count_without_leaf_subsplits_},
+      topology_count_{host_dag.topology_count_},
+      topology_count_below_{host_dag.topology_count_below_} {}
+
+void SubsplitDAG::ResetHostDAG(SubsplitDAG &host_dag) {
+  storage_.ResetHost(host_dag.storage_);
+  dag_taxa_ = host_dag.dag_taxa_;
+  subsplit_to_id_ = host_dag.subsplit_to_id_;
+  parent_to_child_range_ = host_dag.parent_to_child_range_;
+  taxon_count_ = host_dag.taxon_count_;
+  edge_count_without_leaf_subsplits_ = host_dag.edge_count_without_leaf_subsplits_;
+  topology_count_ = host_dag.topology_count_;
+  topology_count_below_ = host_dag.topology_count_below_;
+}
+
 // ** Comparator
 
 int SubsplitDAG::Compare(const SubsplitDAG &other) {
@@ -283,6 +304,10 @@ MutableSubsplitDAGNode SubsplitDAG::GetDAGNode(const size_t node_id) {
 
 size_t SubsplitDAG::GetDAGNodeId(const Bitset &subsplit) const {
   Assert(ContainsNode(subsplit), "Node with the given subsplit does not exist in DAG.");
+  if (storage_.HaveHost()) {
+    auto node = storage_.FindVertex(subsplit);
+    return node->get().GetId();
+  }
   return subsplit_to_id_.at(subsplit);
 }
 
@@ -918,6 +943,9 @@ bool SubsplitDAG::ContainsTaxon(const std::string &name) const {
 }
 
 bool SubsplitDAG::ContainsNode(const Bitset &subsplit) const {
+  if (storage_.HaveHost()) {
+    if (storage_.FindVertex(subsplit).has_value()) return true;
+  }
   return subsplit_to_id_.find(subsplit) != subsplit_to_id_.end();
 }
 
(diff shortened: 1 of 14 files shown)


davidrich27:
I tried to write a new test for graftDAG's.  I was running into issues when trying `AddNodePair` back-to-back.  I think there is an conflict in the SubsplitDAG `AddNodePair` logic.  I think the old `GraftNodePair` had a little less logic, since it wasn't trying to reorder nodes into a topological ordering, it was just appending.

ognian-:
Yes, please push the test. SubsplitDAG::AddNodePair has been changed to only reorder nodes if not in graft mode, but there can be some other issue.

davidrich27:
Okay, I'm getting another bug.  I think the issue may lie with `HavePrepend`.  After `RemoveAllGrafts`, this flag get set false, but we still want to treat `AddNodePair` as a GraftDAG operation.

ognian-:
I pushed the fix, the test is passing

davidrich27 (commented):
Everything looks good and functions correctly on my end!

src/subsplit_dag_storage.hpp
```diff
@@ -378,23 +378,84 @@ class GenericVerticesView {
...
 using ConstVerticesView = GenericVerticesView<const SubsplitDAGStorage>;
 
+template <typename T>
+class PrependableVector {
+public:
+  explicit PrependableVector(PrependableVector<T>* prepend = nullptr) : prepend_{prepend ? &prepend->data_ : nullptr} {}
+
+  T& at(size_t i) {
+    if (!prepend_) return data_.at(i);
+    if (i < prepend_->size()) return (*prepend_)[i];
```
matsen, line 388: Here we have square brackets in an `.at` implementation. Intentional?

src/subsplit_dag_storage.hpp
```diff
@@ -378,23 +378,84 @@ class GenericVerticesView {
 using VerticesView = GenericVerticesView<SubsplitDAGStorage>;
 using ConstVerticesView = GenericVerticesView<const SubsplitDAGStorage>;
 
+template <typename T>
```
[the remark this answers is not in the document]
matsen: What are our expectations about lifetime of the object that this pointer points to?

src/subsplit_dag.cpp
```diff
@@ -1095,18 +1116,23 @@ SubsplitDAG::ModificationResult SubsplitDAG::AddNodePair(const Bitset &parent_su
...
-  edge_reindexer = BuildEdgeReindexer(prev_edge_count);
-  // Update the ids in added_node_ids and added_edge_idxs according to the reindexers.
-  Reindexer::RemapIdVector(added_node_ids, node_reindexer);
-  Reindexer::RemapIdVector(added_edge_idxs, edge_reindexer);
-  // Update fields in the Subsplit DAG according to the reindexers.
-  RemapNodeIds(node_reindexer);
-  RemapEdgeIdxs(edge_reindexer);
-  // Recount topologies to update `topology_count_below_`.
-  CountTopologies();
+  if (!storage_.HavePrepend()) {
```
matsen, line 1119: So we don't reindex at all if we are in the graft DAG setting. This should at least be documented.

src/subsplit_dag.cpp
```diff
@@ -38,6 +38,27 @@ SubsplitDAG::SubsplitDAG(size_t taxon_count,
          "Taxon ids do not cover all indices from 0 to taxon_count.");
 }
 
+SubsplitDAG::SubsplitDAG(SubsplitDAG &prepend, PrependTag)
+    : storage_{prepend.storage_, PrependTag{}},
```
matsen, line 42: If you want to make a move to start using brace initialization, say so, but this isn't our current guideline setup.

matsen (changes requested):
It's awesome to see such a big negative diff!

I was fairly confused at first about "Prepend", which is a verb. Furthermore, I feel like this word only makes sense in the context of the storage itself and it doesn't make sense to have it leak into the more public subsplit DAG API.

Isn't it really a host dag in some sense? What are our expectations about this host dag? What happens when it gets modified while we are using a graft DAG? What can we expect to have work with the graft DAG? Clearly not renumbering. 

These things need to be made more explicit. Thanks!

src/subsplit_dag_storage.hpp
ognian-: Yes it's intentional, because bounds check has already been done in the if() statement

src/subsplit_dag.cpp
ognian-: Correct, that's the behavior of the previous GraftDAG implementation as well. Reindexing would modify the host dag.

src/subsplit_dag.cpp
```diff
@@ -38,6 +38,27 @@ SubsplitDAG::SubsplitDAG(size_t taxon_count,
          "Taxon ids do not cover all indices from 0 to taxon_count.");
 }
 
+SubsplitDAG::SubsplitDAG(SubsplitDAG &prepend, PrependTag)
```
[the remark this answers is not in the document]
ognian-: Correct, I'll change it.

src/subsplit_dag.cpp
ognian-: Sorry, will change that.

src/subsplit_dag_storage.hpp
```diff
@@ -378,23 +378,84 @@ class GenericVerticesView {
...
+    if (!prepend_) return 0;
+    return prepend_->size();
+  }
+
+private:
+  std::vector<T> data_;
+  std::vector<T>* prepend_;
+};
+
+struct PrependTag {};
```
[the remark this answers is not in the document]
ognian-: It's a common technique called "tag dispatching", but I can change terminology to "hint" for example. Without tag dispatching the constructor looks a lot like the copy constructor, and I'll have to define copy ctors and assignment operators to SubsplitDAG and all classes that derives from it. This would be bug prone as well as verbose.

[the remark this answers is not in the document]
matsen: Cool to learn about this, and got it!

Perhaps `PrependDispatchTag` with a little comment and perhaps a link to a description of this idiom?

src/subsplit_dag_storage.hpp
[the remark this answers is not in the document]
matsen: @davidrich27 pointed out that this hasn't been changed from the GraftDAG but let's add some docs, please.

ognian-:
You're right about "prepend" - it shouldn't be used outside storage, I'll correct that. In terms of what's expected to work with GraftDAG - it's meant to be used in NNI code, so mainly storing and querying nodes/edges and comparing DAGs. If the host dag get modified elsewhere the GraftDAG would break, but that's the same behavior as with the previous implementation.

davidrich27:
My intended use case for the host DAG is this:
- First, add modifications the graft DAG.
- After computing NNIs, we will remove all modifications to the graft DAG.
- Finally, we will add permanent modifications to the host DAG.

So changes to the host DAG should only ever occur while the graft DAG is empty.  But the graft DAG does not directly own the host DAG, so it depends on the user not modifying the DAG.

davidrich27:
I added another test to the graft doctest.  Modifying the host DAG after initializing graft DAG is creating issues.

matsen (approved):
Thanks!

PR #420, opened 2022-03-17
Github Action: Run Tests fixes.

### LINKED ISSUE #419: Solve issues with Github Action CI
Github Action CI `Run test` still fails on run.  Passes on Quokka and macOS environments.

davidrich27:
## Description

This fixes the Github Action `Run Tests`.  

*  Selects a version of GCC (uses ver. 7.XX)
*  Symlinks PyTest installed via Apt.

Closes #419 

## Tests

N/A

## Checklist:

### DIFF
--- a/.github/workflows/run-tests-with-gcc-7.yml
+++ b/.github/workflows/run-tests-with-gcc-7.yml
@@ -1,4 +1,4 @@
-name: Run tests
+name: Run tests with GCC-7
 on:
   workflow_dispatch:
 jobs:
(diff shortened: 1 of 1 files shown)


davidrich27:
This is intended as a stopgap, so we have a CI that runs doctest and pytest successfully.  Our tests will eventually need to run correctly on the latest compiler, but for now this runs the CI tests on gcc 7.XX.

ognian-:
Looks good! Is there a chance the GCC from "setup-gcc" be replaced in the next step by "apt install build-essential"?

davidrich27:
Yes.  I was trying to get access to an older version of gcc/g++.  The only version of build-essential my local machine showed was v12.  Now instead of the separate action, it just installs through apt and set an environmental variable.   Is this what you were thinking?

ognian-:
I was worried whether apt install would override the previously installed gcc by setup-gcc. Both versions will be available, but I wasn't sure which one would be the default.

matsen (approved):

PR #426, opened 2022-03-29
Sampling conditional on a subsplit #304

### LINKED ISSUE #304: Sampling conditional on a subsplit
As a first step for local-EM, I'd like to be able to sample rootward from a subsplit, like so:

conditional-topology

As described in the tex, we can do this by using Bayes inversion on the subsplit probabilities. We already have unconditional subsplit probabilities implemented.

ognian-:
## Description

Added a new class (TopologySampler) that is able to sample a random tree out of a SubsplitDAG.

Closes #304 

## Tests

Two tests cover the functionality:
- "TopologySampler"
- "TopologySampler: Non-uniform prior"

## Checklist:

### DIFF
--- a/src/subsplit_dag_storage.hpp
+++ b/src/subsplit_dag_storage.hpp
@@ -267,6 +267,16 @@ class DAGVertex {
     return neighbors_.at({direction, clade});
   }
 
+  bool IsRoot() const {
+    return GetNeighbors(Direction::Rootward, Clade::Left).empty() &&
+           GetNeighbors(Direction::Rootward, Clade::Right).empty();
+  }
+
+  bool IsLeaf() const {
+    return GetNeighbors(Direction::Leafward, Clade::Left).empty() &&
+           GetNeighbors(Direction::Leafward, Clade::Right).empty();
+  }
+
   std::optional<std::tuple<LineId, Direction, Clade>> FindNeighbor(
       VertexId neighbor) const {
     for (auto i : neighbors_) {
--- a/src/subsplit_dag_storage.hpp
+++ b/src/subsplit_dag_storage.hpp
@@ -40,6 +40,14 @@
 enum class Direction { Rootward, Leafward };
 enum class Clade { Unspecified, Left, Right };
 
+inline Clade Opposite(Clade clade) {
+  switch (clade) {
+  case Clade::Left: return Clade::Right;
+  case Clade::Right: return Clade::Left;
+  default: Failwith("Use of unspecified clade");
+  }
+}
+
 using VertexId = size_t;
 using LineId = size_t;
 static constexpr size_t NoId = std::numeric_limits<size_t>::max();
@@ -181,6 +189,8 @@ class GenericNeighborsView {
     bool operator==(const Iterator& other) const { return i_ == other.i_; }
 
     VertexId operator*() { return i_->first; }
+    
+    VertexId GetNodeId() const { return i_->first; }
     LineId GetEdge() const { return i_->second; }
 
    private:
@@ -289,6 +299,15 @@ class DAGVertex {
       }
     }
     Failwith("Neighbor not found");
+  }
+
+  void ClearNeighbors() {
+    neighbors_ = {
+      {{Direction::Rootward, Clade::Left}, {}},
+      {{Direction::Rootward, Clade::Right}, {}},
+      {{Direction::Leafward, Clade::Left}, {}},
+      {{Direction::Leafward, Clade::Right}, {}},
+    };
   }
 
  private:
@@ -504,6 +523,10 @@ class SubsplitDAGStorage {
     return line;
   }
 
+  const DAGVertex& GetVertex(VertexId id) const {
+    return vertices_.at(id);
+  }
+
   bool ContainsVertex(VertexId id) const {
     if (id >= vertices_.size()) return false;
     return vertices_[id].GetId() != NoId;
@@ -573,6 +596,40 @@ class SubsplitDAGStorage {
     }
     lines_.ResetHost(&host.lines_);
     vertices_.ResetHost(&host.vertices_);
+  }
+
+  void ConnectVertices(LineId lineId) {
+    auto& line = lines_.at(lineId);
+    auto& parent = vertices_.at(line.GetParent());
+    auto& child = vertices_.at(line.GetChild());
+    parent.AddNeighbor(Direction::Leafward, line.GetClade(), child.GetId(), lineId);
+    child.AddNeighbor(Direction::Rootward, line.GetClade(), parent.GetId(), lineId);
+  }
+
+  void ConnectAllVertices() {
+    for (size_t i = 0; i < vertices_.size(); ++i) {
+      vertices_[i].ClearNeighbors();
+    }
+    for (size_t i = 0; i < lines_.size(); ++i) {
+      if (lines_[i].GetId() == NoId) {
+        continue;
+      }
+      ConnectVertices(lines_[i].GetId());
+    }
+  }
+
+  std::optional<std::reference_wrapper<const DAGVertex>> FindRoot() const {
+    for (size_t i = 0; i < vertices_.size(); ++i) {
+      auto& vertex = vertices_[i];
+      if (vertex.GetId() == NoId) {
+        continue;
+      }
+      if (vertex.GetNeighbors(Direction::Rootward, Clade::Left).empty() &&
+          vertex.GetNeighbors(Direction::Rootward, Clade::Right).empty()) {
+            return vertex;
+          }
+    }
+    return std::nullopt;
   }
 
  private:
--- a/src/subsplit_dag_storage.hpp
+++ b/src/subsplit_dag_storage.hpp
@@ -182,22 +182,22 @@ class GenericNeighborsView {
  public:
   class Iterator : public std::iterator<std::forward_iterator_tag, VertexId> {
    public:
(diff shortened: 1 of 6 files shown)


src/topology_sampler.cpp
```diff
@@ -0,0 +1,126 @@
...
+  for (auto parent = right.begin(); parent != right.end(); ++parent)
+    weights[i++] = session.inverted_probabilities_[parent.GetEdge()];
+  std::discrete_distribution<> distribution(weights.begin(), weights.end());
+  i = static_cast<size_t>(distribution(mersenne_twister_.GetGenerator()));
+  if (i < left.size()) {
+    auto parent = left.begin();
+    std::advance(parent, i);
+    return {session.dag_.GetDAGNode(parent.GetNodeId()),
+            session.dag_.GetDAGEdge(parent.GetEdge())};
+  }
```
matsen, line 79:
```suggestion
  } // else
```

I like to put this around early returns

src/topology_sampler.cpp
```diff
@@ -0,0 +1,126 @@
...
+  session.result_.AddLine({child.second.GetId(), child.second.GetParent(),
+                           child.second.GetChild(), child.second.GetClade()});
+  VisitNode(session, child.first, Direction::Rootward, clade);
+}
+
+std::pair<SubsplitDAGNode, ConstLineView> TopologySampler::SampleParent(
+    SamplingSession& session, ConstNeighborsView left, ConstNeighborsView right) {
+  std::vector<double> weights;
+  weights.resize(left.size() + right.size());
+  size_t i = 0;
```
matsen, line 67: Let's talk about `i` with Dave.

src/topology_sampler.hpp
```diff
@@ -0,0 +1,139 @@
...
+
+  // Set a seed value for the internal random number generator.
+  void SetSeed(uint64_t seed);
+
+ private:
+  struct SamplingSession {
+    SubsplitDAG& dag_;
+    EigenConstVectorXdRef normalized_sbn_parameters_;
+    EigenConstVectorXdRef inverted_probabilities_;
+    SubsplitDAGStorage result_;
```
matsen, line 33: Interesting that we're using the `SubsplitDAGStorage` here, which I thought was going to be an "implementation detail" of SubsplitDAG. I'm not saying that we shouldn't, but it's worth thinking about how much this is a core data structure in the code.

src/topology_sampler.hpp
```diff
@@ -0,0 +1,139 @@
...
+      dag.UnconditionalNodeProbabilities(normalized_sbn_parameters);
+  EigenVectorXd inverted_probabilities =
+      dag.InvertedGPCSPProbabilities(normalized_sbn_parameters, node_probabilities);
+
+  SubsplitDAGNode origin = dag.GetDAGNode(5);
+
+  TopologySampler sampler;
+  std::map<std::string, size_t> counts;
+  std::map<std::string, double> expected = {
+      {"((((x0,x1),x2),(x3,x4)));", 0.312},
```
matsen, line 118: Could we get these numbers "on the fly" by sampling leafward?

src/topology_sampler.hpp
```diff
@@ -0,0 +1,139 @@
...
+  for (size_t i = 0; i < iterations; ++i) {
+    auto tree =
+        sampler.Sample(origin, dag, normalized_sbn_parameters, inverted_probabilities);
+    ++counts[tree->Newick([](const Node* node) {
+      if (!node->IsLeaf()) return std::string();
+      return std::string("x") + std::to_string(node->Id());
+    })];
+  }
+
+  for (auto& i : counts) {
```
matsen, line 133: structured binding, please!

src/topology_sampler.hpp
```diff
@@ -0,0 +1,139 @@
...
+    EigenConstVectorXdRef normalized_sbn_parameters_;
+    EigenConstVectorXdRef inverted_probabilities_;
+    SubsplitDAGStorage result_;
+  };
+
+  // Called for each newly sampled node. Direction and clade are pointing to the
+  // previously visited node. 
+  void VisitNode(SamplingSession& session, SubsplitDAGNode node, Direction direction,
+                 Clade clade);
+  // Continue sampling in the rootward direction from "node".
```
matsen, line 40:
```suggestion
  // Continue sampling in the rootward direction from `node`.
```

This is my preference when describing a named object.

src/topology_sampler.cpp
```diff
@@ -0,0 +1,126 @@
...
+    // reached leaf
+    return;
+  }
+  auto child = SampleChild(session, neighbors);
+  session.result_.AddLine({child.second.GetId(), child.second.GetParent(),
+                           child.second.GetChild(), child.second.GetClade()});
+  VisitNode(session, child.first, Direction::Rootward, clade);
+}
+
+std::pair<SubsplitDAGNode, ConstLineView> TopologySampler::SampleParent(
```
matsen, line 63:
```suggestion
std::pair<SubsplitDAGNode, ConstLineView> TopologySampler::SampleParentNodeAndEdge(
```

src/subsplit_dag_storage.hpp
```diff
@@ -181,6 +192,8 @@ class GenericNeighborsView {
     bool operator==(const Iterator& other) const { return i_ == other.i_; }
 
     VertexId operator*() { return i_->first; }
+
+    VertexId GetNodeId() const { return i_->first; }
```
matsen, line 196: Ha! I clearly didn't look this over very closely on the first round. Having a member variable named `i_` clearly isn't enough to know what it is. `iter_`?

src/topology_sampler.hpp
matsen: Note that this is a question, and not a TODO necessarily!

matsen (changes requested):
Very neat! I look forward to learning more about this tomorrow. A few things to fix/discuss here.

src/subsplit_dag_storage.hpp
```diff
@@ -575,6 +599,40 @@ class SubsplitDAGStorage {
...
+    }
+    for (size_t i = 0; i < lines_.size(); ++i) {
+      if (lines_[i].GetId() == NoId) {
+        continue;
+      }
+      ConnectVertices(lines_[i].GetId());
+    }
+  }
+
+  std::optional<std::reference_wrapper<const DAGVertex>> FindRoot() const {
```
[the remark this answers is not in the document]
ognian-: Ordering of nodes is a concept of the SubsplitDAG, not the storage layer. There are situations (like in Graft, or when adding nodes) when the ordering is yet to be enforced. For users of SubsplitDAG ordering looks like an atomic operation - there is no call in the sDAG interface that would leave it in a state where the root isn't the last node, but during internal usage we don't have such guarantee. The function is called FindRoot instead of GetRoot to denote higher complexity than O(1).

src/topology_sampler.cpp
```diff
@@ -0,0 +1,126 @@
+// Copyright 2019-2021 bito project contributors.
+// bito is free software under the GPLv3; see LICENSE file for details.
+
+#include "topology_sampler.hpp"
+
+Node::NodePtr TopologySampler::Sample(SubsplitDAGNode node, SubsplitDAG& dag,
+                                      EigenConstVectorXdRef normalized_sbn_parameters,
+                                      EigenConstVectorXdRef inverted_probabilities) {
+  SamplingSession session{dag, normalized_sbn_parameters, inverted_probabilities};
```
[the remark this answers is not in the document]
ognian-: To fully remove the brace initialization, I'll have to add a constructor to SamplingSession. For now I'm changing it to:

`SamplingSession session({dag, normalized_sbn_parameters, inverted_probabilities});`

And we can discuss further.

src/topology_sampler.cpp
```diff
@@ -0,0 +1,126 @@
...
+    weights[i++] = session.normalized_sbn_parameters_[child.GetEdge()];
+  std::discrete_distribution<> distribution(weights.begin(), weights.end());
+  i = static_cast<size_t>(distribution(mersenne_twister_.GetGenerator()));
+  auto child = neighbors.begin();
+  std::advance(child, i);
+  return {session.dag_.GetDAGNode(child.GetNodeId()),
+          session.dag_.GetDAGEdge(child.GetEdge())};
+}
+
+Node::NodePtr TopologySampler::BuildTree(SamplingSession& session,
```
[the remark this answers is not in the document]
ognian-: It's my take on what we discussed about having to build the Node tree starting from the root. The function is expecting that each node in` session.result_` has at most one child in left and right clades, which is exactly the output of the rest of the functions in TopologySampler. I wasn't sure whether we always get fully balanced trees with two children on every node, if that's the case I can replace the two checks `if (left_id != NoId)` and `if (right_id != NoId)` with assertions.

src/topology_sampler.hpp
ognian-: It's not essential as it is only used to build the resulting Node tree, but it reduces the complexity in traversing the sampled nodes. A heavier alternative would be to use a throwaway instance of SubsplitDAG, or using a map with <Node, pair<Node, Node> for the two children of each node. I'm leaving it unchanged for further discussion.

src/topology_sampler.hpp
ognian-: Yes, but it would increase the runtime duration of the test, which is already somewhat lengthy.

matsen (approved):
Looks great!

PR #449, opened 2022-06-17
Top-Pruning: ChoiceMap, pt2

### LINKED ISSUE #447: Top-Pruning: ChoiceMap -- TreeMask Update and Build Trees
Currently, our `ChoiceMap` has a method for building a `TreeMask`, which is a subset of the edges in the SubsplitDAG.  This is currently a vector of edge ids, but should be updated to a set of edge ids.  Additionally, this will add a function for building a `Tree` from a `ChoiceMap` so that down the road we can test top-pruning trees likelihoods by comparing them against computations using BEAGLE.

davidrich27:
## Description

Makes updates to ChoiceMaps.
-  TreeMasks are now sets instead of vectors of edge ids.
-  Adds functionality for building Topology from ChoiceMaps based on an edge in DAG.
-  Tests for if a given Topology exists in the DAG

Closes #447

## Tests

TEST_CASE("Top-Pruning: ChoiceMap"): Updated to include test for if Topologies exist in the DAG.

## Checklist:


src/choice_map.hpp
```diff
@@ -32,123 +31,24 @@ class ChoiceMap {
...
+  using TreeMask = std::set<size_t>;
+  enum class AdjacentNode { Parent, LeftChild, RightChild };
   template <typename T>
-  static void StackPushIfValidId(std::stack<T> &stack, size_t id) {
-    if (id != NoId) {
-      stack.push(id);
-    }
-  }
+  using AdjacentNodeArray = EnumArray<AdjacentNode, 3, T>;
+  using ExpandedTreeMask = std::map<size_t, AdjacentNodeArray<size_t>>;
```
matsen, line 51: I'm just noting that if we had a NodeId type this typedef would be a bunch clearer!

src/choice_map.hpp
```diff
@@ -158,154 +58,49 @@ class ChoiceMap {
...
+  std::string TreeMaskToString(const TreeMask &tree_mask) const;
+  std::string ExpandedTreeMaskToString(const ExpandedTreeMask &tree_mask) const;
 
-    return true;
-  }
+  // ** Tree
+
+  // Extract Tree from DAG based on edge choices to find best tree with given
+  // central edge.
+  // - Makes two passes:
```
matsen, line 85: These are implementation docs, which I love, but they belong in the cpp file.

src/gp_doctest.cpp
```diff
@@ -1567,9 +1568,11 @@ TEST_CASE("NNI Engine: NNI Likelihoods") {
 // listed neighbor for each parent, sister, left and right child. Tests that results is
 // a valid selection (all edges have mapped valid edge choices, except for root and
 // leaves). Then creates TreeMasks for each edge in DAG, a list of edge ids which
-// represent a embedded tree in the DAG.  Tests that each TreeMask contains its central
-// edge and is valid (tree spans root and all leaf nodes, and every node in tree has a
-// single parent, left child and right child).
+// represent a embedded tree in the DAG.
+// - Tests that invalid TreeMasks and Trees are found in invalid.
```
matsen, line 1572:
```suggestion
// - Tests that invalid TreeMasks and Trees are found to be invalid.
```

src/node.hpp
```diff
@@ -189,6 +189,11 @@ class Node {
   // https://stackoverflow.com/a/776523/467327
   static size_t SORotate(size_t n, uint32_t c);
 
+  // Output single node to string.
```
matsen, line 192: There's a specific sense that this is a node to string, and I wouldn't know that by looking at the header file. Perhaps `NodeIdAndLeavesToString`? And the whole tree version could be `NodeIdAndLeavesToStringForTopology`?

I usually think of a Newick string when I think of a generic "tree to string" thing.

src/choice_map.hpp
```diff
@@ -158,154 +58,49 @@ class ChoiceMap {
...
+  //   - The first pass goes up along the chosen edges of the DAG to the root, adding
+  //   each edge it encounters.
+  //   - The second pass goes leafward, descending along the chosen edges to the leaf
+  //   edges from the sister of each edge in the rootward pass and the child edges from
+  //   the central edge.
+  Tree ExtractTree(const size_t central_edge_id) const;
+  Tree ExtractTree(const TreeMask &tree_mask) const;
+  Tree ExtractTree(ExpandedTreeMask &tree_mask_ext) const;
+
+  // Checks that tree is a valid tree in DAG that spans the
```
matsen, line 95: incomplete

src/choice_map.cpp
```diff
@@ -0,0 +1,442 @@
...
+    Assert(next_node_id != current_node_id, "Node cannot be adjacent to itself.");
+    current_node_id = next_node_id;
+  }
+
+  Node::NodePtr topology = nodes[dag_rootsplit_id];
+  topology->Polish();
+  Assert((nodes.size() + 1) == tree_mask_ext.size(),
+         "Invalid TreeMask-to-Tree: Topology did not span every node in "
+         "the TreeMask.");
+  Tree tree(topology, Tree::BranchLengthVector(nodes.size()));
```
matsen, line 381: I guess this is an uninitialized vector of doubles. Should we set them all to -1 as a warning that they are meaningless or something?

src/choice_map.cpp
```diff
@@ -0,0 +1,442 @@
...
+
+  return tree;
+}
+
+bool ChoiceMap::TreeIsValid(const Tree &tree, const bool is_quiet) const {
+  std::ofstream devnull("/dev/null");
```
matsen, line 387: Elsewhere (mmapped_matrix) I just used a stringstream that gets thrown away. Look for `dev_null`. Perhaps I prefer that a little bit because it's not making a kernel call. You?

src/choice_map.cpp
matsen: Or should we really just be dealing with topologies here?

src/choice_map.cpp
```diff
@@ -0,0 +1,442 @@
...
+  BoolVector visited_leaves(leaves.size(), false);
+  if (leaves.size() != dag_.TaxonCount()) {
+    os << "Invalid Tree: tree does not cover all taxa.";
+    return false;
+  }
+  if (!leaves.All()) {
+    os << "Invalid Tree: root does not span all taxa.";
+    return false;
+  }
+  tree.Topology()->DepthFirst(
```
matsen, line 401: Isn't this just a preorder traversal? I think I saw this somewhere else too.

src/choice_map.cpp
```diff
@@ -0,0 +1,442 @@
...
+  topology->Polish();
+  Assert((nodes.size() + 1) == tree_mask_ext.size(),
+         "Invalid TreeMask-to-Tree: Topology did not span every node in "
+         "the TreeMask.");
+  Tree tree(topology, Tree::BranchLengthVector(nodes.size()));
+
+  return tree;
+}
+
+bool ChoiceMap::TreeIsValid(const Tree &tree, const bool is_quiet) const {
```
matsen, line 386: Interesting to have this here... it's not about a choice map at all. I think that you are building a tree by specifying leaves and joining them... is it possible to make an invalid tree? If so we should think about the node API and perhaps some tests over there.

src/gp_doctest.cpp
```diff
@@ -1584,78 +1587,94 @@ TEST_CASE("Top-Pruning: ChoiceMap") {
...
+  CHECK_FALSE_MESSAGE(choice_map.TreeMaskIsValid(tree_mask, quiet_errors),
                       "TreeMask is incorrectly valid when missing internal edge.");
+  CHECK_THROWS_MESSAGE(choice_map.ExtractTree(tree_mask),
+                       "Tree is incorrectly valid when missing internal edge.");
   // Tree contains additional edge.
   tree_mask = choice_map.ExtractTreeMask(0);
-  for (size_t i = 0; i < dag.EdgeCountWithLeafSubsplits(); i++) {
-    const auto contains_edge =
-        (std::find(tree_mask.begin(), tree_mask.end(), i) != tree_mask.end());
+  tree = choice_map.ExtractTree(tree_mask);
```
matsen, line 1648: Do we use this tree?

matsen (changes requested):
Nice, Dave! A few little things here.

Is there a test for ExtractTree? I'm not sure if I see sort of a complete test for the functionality.

src/choice_map.hpp
davidrich27: Yes, we need to make `NodeId` and `EdgeId` strongly typed across Bito.  We already have an issue raised for that.  I'll work on that this week!

src/gp_doctest.cpp
davidrich27: Fixed.

src/choice_map.hpp
davidrich27: Fixed.

src/choice_map.hpp
davidrich27: Moved.

src/choice_map.cpp
davidrich27: I think these branch lengths can be in logscale, right?  So would probably need to be a NaN or something.

Regardless, I think its just cleaner to make them topologies.

src/choice_map.hpp
```diff
@@ -158,154 +58,49 @@ class ChoiceMap {
...
-      leafward_stack.pop();
-      tree_mask.push_back(edge_id);
-      const auto edge_choice = edge_choice_vector_.at(edge_id);
-      StackPushIfValidId(leafward_stack, edge_choice.left_child_edge_id);
-      StackPushIfValidId(leafward_stack, edge_choice.right_child_edge_id);
-    }
-
-    return tree_mask;
-  }
+  // Expand a TreeMask into a ExpandedTreeMask.  Rather than containing all the edges in
```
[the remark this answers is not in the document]
davidrich27: Moved.

src/choice_map.cpp
davidrich27: Yep.  Changed to Preorder.

src/choice_map.cpp
davidrich27: It is possible to make an invalid tree by my constraints.  It needs to include all the leaf nodes in the DAG.  And it needs to strictly bifurcating (`NodePtr` topologies allows multi-furcations).  Node throw errors during construction through joins for the other types of invalid topologies, like edges that terminate at non-leaf nodes and duplicate leaves.

src/choice_map.cpp
davidrich27: You're right, that is cleaner.  Switched it.

src/choice_map.cpp
davidrich27: As far as where it should go, you could have it in node, but you would also need to pass it a DAG to check against for validity.  Because of that, maybe SubsplitDAG is the best place for it?

matsen: It seems to me that it only uses the taxon count from the DAG. Couldn't we just have that be an argument? 

If we put it in SubsplitDAG, I would have thought that it would be checking that the tree could be generated from the DAG, which this doesn't do. OTOH such a function could be helpful.

We should also note that this checks that a tree is bifurcating... it's "valid" to have a non-bifurcating tree though not in the context of the DAG.

src/choice_map.cpp
matsen: Great. Thanks for making it clean.

matsen (commented):
Do we actually have a test to make sure that ExtractTopology is doing the _right_ thing?

src/choice_map.cpp
davidrich27: Right.  That is the only argument we need in the taxon count.

As far as testing whether ExtractTopology is doing the right thing, the ChoiceMap selection method is entirely naive, just selecting the "first" parent, left and right child for each node in the DAG.  The nature of this "first" neighbor is subject to the implementation of the DAG and even the ordering of the input, if generated from a Newick tree file.  

So it feels like if the topology is valid and could be generated from the DAG (I agree we should have that test), then we should accept the topology as correct.  I think when we have more deterministic ChoiceMap selection methods, we can have more concrete testing in this regard.

src/gp_doctest.cpp
```diff
@@ -1584,78 +1587,141 @@ TEST_CASE("Top-Pruning: ChoiceMap") {
...
       }
     }
   }
-  CHECK_FALSE_MESSAGE(choice_map.TreeMaskIsValid(tree_mask),
+  CHECK_FALSE_MESSAGE(choice_map.TreeMaskIsValid(tree_mask, quiet_errors),
                       "TreeMask is incorrectly valid when containing an extra edge.");
+  CHECK_THROWS_MESSAGE(choice_map.ExtractTopology(tree_mask),
+                       "Tree is incorrectly valid when containing an extra edge.");
+
+  auto NodeLeaf = [](const size_t id, const size_t taxon_count) -> Node::NodePtr {
```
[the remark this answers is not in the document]
matsen: But why do we need an explicit bitset here and not below?

src/choice_map.cpp
matsen: OK, do we need topology validity testing if we check that it exists in the DAG? And are you good regarding how to test if something sits in the DAG or not?

src/gp_doctest.cpp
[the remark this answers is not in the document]
davidrich27: Sorry, it should be everywhere.  I originally was populating the `leaves` clade for the nodes with `Node::Polish()`, but in that case the leaves are overwritten according to the order they are visited in the postorder traversal.  So I switched over to using explicit bitsets.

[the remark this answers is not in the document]
matsen: OK, if this is generally useful then don't hesitate to make an actual `node.hpp` function out of it.

src/choice_map.cpp
davidrich27: I think the test if it exists in the DAG should be sufficient without a Topology validation test. My thought for this "topology exists in DAG" test would be to check via PCSPs.  Check if all PCSPs in the topology are also in the DAG.  I think we would include verifying that topology reaches every leaf node in this test and eliminate the valid topology test.

matsen: Great!

src/gp_doctest.cpp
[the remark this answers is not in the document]
davidrich27: Added function

src/subsplit_dag.cpp
```diff
@@ -998,6 +998,80 @@ bool SubsplitDAG::IsEdgeLeaf(const size_t edge_id) const {
...
+        os << "DoesNotContainTopology: Non-leaf node does not have 2 children."
+           << std::endl;
+        contains_topology = false;
+        return;
+      }
+      const auto parent_subsplit =
+          Bitset::Subsplit(child_nodes[0]->Leaves(), child_nodes[1]->Leaves());
+      for (const auto child_node : child_nodes) {
+        const auto grandchild_nodes = child_node->Children();
+        // Build child subsplit based on if it is a .
```
matsen, line 1045: Incomplete comment

matsen (approved):
Just noted one comment that needs fixing, but otherwise 👍 .

PR #452, opened 2022-06-30
PLVHandler Update

### LINKED ISSUE #446: Top-Pruning: PLVHandler Update
We have a current implementation of the PLVHandler that is a collection of static functions for the GPEngine's PLV vectors.  This issue will move ownership of the PLV vector to a PLVHandler instance, so that it can be used by our TopEngine(?).  Additionally, the number of PLVTypes is currently hard-coded as 6.  This will add the option for a 2 PLVType PLVHandler.

davidrich27:
## Description

PVHandler (previously PLVHandler)
- Moved data for PLVs from GPEngine to PVHandler.
- Moved all data handling (init, resize, reindex) to PLVHandler.
- Added templating to support different sets, for algorithms using partials different from Generalized Pruning.

Closes #446 

## Tests

N/A.  Previous tests from GPEngine ensure proper functionality.

## Checklist:


davidrich27:
@matsen Hey Erick, I was wondering if you have an opinion on the naming conventions for the `PLVHandler`?  Since its now generalized beyond just GP PLVs, including different kinds of cost matrices.  Would you refer to them generally as Partial Vectors?

src/gp_engine.cpp
```diff
@@ -17,11 +15,8 @@ GPEngine::GPEngine(SitePattern site_pattern, size_t node_count,
     : site_pattern_(std::move(site_pattern)),
       rescaling_threshold_(rescaling_threshold),
       log_rescaling_threshold_(log(rescaling_threshold)),
-      plv_count_per_node_(plv_count_per_node),
-      mmap_file_path_(mmap_file_path),
-      mmapped_master_plv_(mmap_file_path_,
-                          (node_count + node_padding_) * plv_count_per_node_ *
-                              size_t(resizing_factor_) * site_pattern_.PatternCount()),
+      plv_handler_(mmap_file_path, size_t(0), site_pattern_.PatternCount(),
```
matsen, line 18: Is this a cast of 0 to `size_t`? Is it needed?

matsen (changes requested):
I was confused for a bit about the presence of `padding` and `Temp`, which are referring to the same thing, though I don't think that either name is really descriptive. "padding" is what you do if you need something of a certain shape, and "Temp" I'd think was actually temporary. This is more like "utility" or "spare". What do you think?

src/gp_engine.cpp
davidrich27: Nope, sorry.  That was from a templating issue I was having.  Fixed.

davidrich27:
Spare like a good name for that to me.  Made name change.

PR #454, opened 2022-07-05
Sankoff on trees

### LINKED ISSUE #448: parsimony 1: perform Sankoff on trees
Here's a place to discuss data structures for Sankoff in bito.

The first task is to calculate Sankoff on a tree (i.e. a rooted tree represented by a Node object). 

Let's plan ahead so that we will be able to use this in the context of tree/DAG modifications. Specifically, we will want to have a data structure that represents the matrix used in Sankoff calculation.

I propose that we have two objects to drive this work.

* A `SankoffHandler` that keeps track of everything needed to do Sankoff 
* A `SankoffMatrix` that represents the partial computation as part of the Sankoff algorithm

Notes:

* The constructor for the `SankoffHandler` should take a [SitePattern](https://github.com/phylovi/bito/blob/main/src/site_pattern.hpp)
* The `SankoffHandler` should be able to generate a matrix for each tip of the tree
* It should also have a method that, given two matrices for child nodes, calculates the matrix for the parent node
* It should also have a method that can take a tree and do a [postorder traversal](https://github.com/phylovi/bito/blob/main/src/node.hpp#L94) to calculate the Sankoff score.

tanviganapathy:
## Description

New objects to calculate the parsimony score of a topology using the Sankoff algorithm. 
- `SankoffMatrix` represents the 4x4 matrix representing mutation costs. 
- `SankoffHandler` takes in a `SitePattern` and `SankoffMatrix` and contains all methods to run Sankoff on a topology.
- `RunSankoff` method in `SankoffHandler` calculates p-partials and q-partials for the topology.
- `ParsimonyScore` method in `SankoffHandler` uses p-partials and q-partials to find the total parsimony score of topology.

Closes #448 

## Tests

Tests added to both `SankoffMatrix` and `SankoffHandler` classes.


src/sankoff_matrix.hpp
```diff
@@ -0,0 +1,71 @@
+// Copyright 2019-2022 bito project contributors.
```
matsen: @ognian- could you have a look over this file?

src/sankoff_handler.cpp
```diff
@@ -0,0 +1,105 @@
+// Copyright 2019-2022 bito project contributors.
+// bito is free software under the GPLv3; see LICENSE file for details.
+
+#include "sankoff_handler.hpp"
+
+Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count> SankoffHandler::GenerateLeavesPartials(const size_t node_count, const size_t pattern_idx) {
```
matsen: Perhaps "GenerateLeafPartials"? Also ccing @davidrich27 to note that here you're just saying "partials" rather than "partial vectors".

src/sankoff_handler.cpp
```diff
@@ -0,0 +1,105 @@
...
+    auto site_partials = Eigen::Matrix<double, Eigen::Dynamic, state_count>();
+    site_partials.resize(node_count, state_count);
+    site_partials.setZero();
+    for (size_t seq = 0; seq < site_pattern_.TaxonCount(); seq++) {
+        auto site_val = site_pattern_.GetPatternSymbol(seq, pattern_idx);
+        if (site_val < state_count) {
+            for (size_t i = 0; i < state_count; i++) {
+                if (i != site_val) {
+                    site_partials(seq, i) = static_cast<double>(INT_MAX);
```
matsen: I was at first confused at the mix of ints and doubles here, but I guess it makes sense because we don't want to overflow a double, but have a big number.

src/sankoff_matrix.hpp
matsen: I note that if we template on state_count we can use the fixed-size Eigen matrices, which have advantages for efficiency.

src/sankoff_handler.cpp
matsen: In any case can we have this be a class-level const? Perhaps it could be constexpr? It shows up a bunch in the tests so let's change it there.

src/sankoff_handler.cpp
```diff
@@ -0,0 +1,105 @@
...
+                    q_partials[pattern_idx].block<1, state_count>(node->Children()[child]->Id(), 0) = partials_from_sister + partials_from_parent;
+                }
+            }
+        } 
+    });
+
+    return {p_partials_left, p_partials_right, q_partials};
+} 
+
+double SankoffHandler::ParsimonyScore(std::tuple<std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>, std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>, std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>> all_partials, size_t node_id) {
```
matsen: Why do a tuple here rather than just pass the three named things? 

Would a typedef clean this up?

Are these constant objects? Could we pass references?

src/sankoff_handler.hpp
```diff
@@ -0,0 +1,127 @@
...
+#include "driver.hpp"
+
+class SankoffHandler {
+    public:
+        // DNA assumption
+        static constexpr size_t state_count = 4;
+        // bifurating tree assumption
+        static const size_t child_count = 2;
+
+    SankoffHandler(SitePattern site_pattern) {
```
matsen: C++ initializer list would be more common for a header.

matsen (changes requested):
After reformatting I can't see or edit my comments any more, but hopefully they still make sense. 

On second thought I'm not sure that we should template on the state count because then everything will have to be templated.

In any case I'm submitting this now so I can see my comments and then will comment more tomorrow.

src/sankoff_handler.cpp
```diff
@@ -3,23 +3,28 @@
...
-    auto site_val = site_pattern_.GetPatternSymbol(seq, pattern_idx);
-    if (site_val < state_count) {
-      for (size_t i = 0; i < state_count; i++) {
-        if (i != site_val) {
-          site_partials(seq, i) = static_cast<double>(INT_MAX);
+SankoffPartialVec SankoffHandler::GenerateLeafPartials(const size_t node_count) {
+  SankoffPartialVec p_partials_left(site_pattern_.PatternCount());
+  for (size_t pattern_idx = 0; pattern_idx < site_pattern_.PatternCount();
+       pattern_idx++) {
+    // p_partials_left[pattern_idx] =
```
matsen, line 10: stray code?

src/sankoff_handler.cpp
```diff
@@ -3,103 +3,158 @@
...
-    });
+SankoffPartialVec SankoffHandler::RunSankoff(Node::NodePtr topology) {
+  // generating p_partials (right and left)
+  SankoffPartialVec p_partials_left = GenerateLeafPartials(topology->Id() + 1);
+  auto p_partials_right = BlankPartials(topology->Id() + 1);
+  topology->Postorder([&p_partials_left, &p_partials_right, this](const Node* node) {
+    if (!node->IsLeaf()) {
+      for (size_t pattern_idx = 0; pattern_idx < site_pattern_.PatternCount();
+           pattern_idx++) {
+        Assert(node->Children().size() == child_count,
```
matsen, line 85: `child_count` is a variable and the error is about exactly 2.

src/sankoff_handler.cpp
```diff
@@ -3,103 +3,158 @@
...
+    }
+  });
 
-double SankoffHandler::ParsimonyScore(std::tuple<std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>, std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>, std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>> all_partials, size_t node_id) {
-    // all_partials: {p_partials_left, p_partials_right, q_partials}
+  // Compress the p_partials_left, p_partials_right, and q_partials for one site
+  // together and use enum SankoffPartialType to separate the types.
+  auto concatenated_partials = BlankPartials(3 * (topology->Id() + 1));
+  for (size_t pattern = 0; pattern < site_pattern_.PatternCount(); pattern++) {
+    concatenated_partials[pattern] << p_partials_left[pattern],
```
matsen, line 130: Whoa, is this Eigen magic?

src/sankoff_handler.hpp
```diff
@@ -12,116 +12,159 @@
...
-    std::tuple<std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>, std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>, std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>> RunSankoff(Node::NodePtr topology);
-
-    // Calculates parsimony score on given node across all sites.
-    double ParsimonyScore(std::tuple<std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>, std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>, std::vector<Eigen::Matrix<double, Eigen::Dynamic, SankoffHandler::state_count>>> all_partials, size_t node_id);
-
-  private:
-    SankoffMatrix mutation_costs_;
-    SitePattern site_pattern_;
+ public:
+  static inline const size_t partial_type_count = 3;
```
matsen, line 22: All of these members should end with an underscore (sorry to miss this the first time around).

src/sankoff_handler.cpp
```diff
@@ -3,103 +3,158 @@
...
-        auto site_val = site_pattern_.GetPatternSymbol(seq, pattern_idx);
-        if (site_val < state_count) {
-            for (size_t i = 0; i < state_count; i++) {
-                if (i != site_val) {
-                    site_partials(seq, i) = static_cast<double>(INT_MAX);
-                }
-            }
+      auto site_val = site_pattern_.GetPatternSymbol(seq, pattern_idx);
+      if (site_val < state_count) {
+        for (size_t i = 0; i < state_count; i++) {
```
matsen, line 17: This is so not important, but one could set everything to a constant `big_double` value and then come through and just change the `site_val` entry to 0.

src/sankoff_handler.cpp
```diff
@@ -3,103 +3,158 @@
...
     for (size_t seq = 0; seq < site_pattern_.TaxonCount(); seq++) {
-        auto site_val = site_pattern_.GetPatternSymbol(seq, pattern_idx);
-        if (site_val < state_count) {
-            for (size_t i = 0; i < state_count; i++) {
-                if (i != site_val) {
-                    site_partials(seq, i) = static_cast<double>(INT_MAX);
-                }
-            }
+      auto site_val = site_pattern_.GetPatternSymbol(seq, pattern_idx);
+      if (site_val < state_count) {
```
matsen, line 16: Shouldn't we be raising an error if site_val is out of bounds? Or is this to handle nucleotide ambiguity codes?

tanviganapathy: It's for handling nucleotide ambiguity and gaps in the sequence alignment. I added a comment in the code.

src/sankoff_handler.cpp
tanviganapathy: @matsen I don't understand this comment. child_count is to assert that the tree is bifurcating.

matsen: Right-- I'm saying that it makes more sense to just write 2 because that's what the error message in the assert says. I'm not sure if it makes sense to have child_count at all-- I can't imagine a setting in which we would want a value other than 2.

matsen: (But I appreciate the intent to keep things clear and tidy!)

src/sankoff_handler.cpp
tanviganapathy: I did this to address the issue of passing the tuple around. Now for each site, there is a SankoffPartial where the first 3rd represents the left p-partials, etc.

matsen: I get that, and it's awesome. I was just curious about the syntax here. Is `<<` an Eigen operator?

tanviganapathy: `<<` is the Eigen comma initializer: https://eigen.tuxfamily.org/dox/group__TutorialAdvancedInitialization.html.

marybarker:
Updates to the SankoffHandler class:

- integrated `PSVHandler` instance to keep track of the partial vectors internally, rather than passing into all of the function calls
- refactored the routines and tests to work with the indexing inherited from the 'PSVHandler' object

src/sankoff_handler.cpp
```diff
@@ -3,34 +3,38 @@
...
+         pattern_idx++) {
+      auto site_val = site_pattern_.GetPatternSymbol(leaf_node, pattern_idx);
       if (site_val < state_count_) {
-        site_partials(seq, site_val) = 0.;
+        node_partials(site_val, pattern_idx) = 0.;
       } else if (site_val == state_count_) {
         // Leaves with gaps in sequence and ambiguous nucleotides are assigned sankoff
         // partial vector [0, 0, 0, 0] at the corresponding site.
-        site_partials.block<1, state_count_>(seq, 0).fill(0);
+        node_partials.block<state_count_, 1>(0, pattern_idx).fill(0);
```
matsen, line 28: Given that this sort of statement gets repeated a lot, can we have an attractively-named alternative with zero overhead so we could say something like `GetColumn(node_partials, pattern_idx).fill(0)`.

cc @ognian- who now knows a lot about eigen ref types.

src/sankoff_handler.cpp
```diff
@@ -53,115 +57,92 @@ EigenVectorXd SankoffHandler::ParentPartial(EigenVectorXd child_partials) {
...
-  for (size_t pattern_idx = 0; pattern_idx < site_pattern_.PatternCount();
-       pattern_idx++) {
-    SankoffPartial blank_partials_per_site(node_count, state_count_);
-    blank_partials_per_site.setZero();
-    blank_partials[pattern_idx] = blank_partials_per_site;
-  }
-  return blank_partials;
-}
+void SankoffHandler::RunSankoff(Node::NodePtr topology) {
+  // make sure that the psv_handler is sized to fit the current topology
```
matsen, line 68: I'm not sure if a comment is needed here, but if you want one let's make a full grammatical sentence, please.

src/sankoff_handler.cpp
```diff
@@ -53,115 +57,92 @@ EigenVectorXd SankoffHandler::ParentPartial(EigenVectorXd child_partials) {
...
+  psv_handler_.SetNodeCount(topology->Id() + 1);
+  psv_handler_.SetAllocatedNodeCount(
+		  size_t(ceil(double(
+			psv_handler_.GetPaddedNodeCount()) * resizing_factor_)));
+  psv_handler_.Resize(topology->Id() + 1, psv_handler_.GetAllocatedNodeCount());
+
+  // fill in leaf node partials for PSV (stored in p_left in psv_
```
matsen, line 74: incomplete

src/sankoff_handler.hpp
```diff
@@ -11,81 +11,108 @@
...
-  enum class SankoffPartialType : size_t { PLeft = 0, PRight = 1, Q = 2 };
-  using SankoffTypeIterator =
-      EnumIterator<SankoffPartialType, SankoffPartialType::PLeft,
-                   SankoffPartialType::Q>;
-
   // DNA assumption
   static constexpr size_t state_count_ = 4;
   static constexpr double big_double_ = static_cast<double>(INT_MAX);
+  // this is for the pvhandler 
+  constexpr static double resizing_factor_ = 2.0;
```
matsen, line 30: Perhaps this should stay in the PV handler? cc @davidrich27

src/sankoff_handler.hpp
```diff
@@ -99,45 +126,51 @@ TEST_CASE("SankoffHandler: Tests on single site sequence.") {
...
   Node::NodePtr topology = tree_collection.GetTree(0).Topology();
+  size_t taxon_count = site_pattern.TaxonCount();
 
   // transitions have cost 1 and transversions have cost 2.5
   auto costs = CostMatrix();
   costs << 0., 2.5, 1., 2.5,  //
       2.5, 0., 2.5, 1.,       //
       1., 2.5, 0., 2.5,       //
       2.5, 1., 2.5, 0.;       //
+
```
matsen, line 131: My guess is that things are fine, but it might be nice to have a less "symmetric" test. You might have to ask Will D for help running Sankoff with a different matrix.

matsen (commented):
Great! I just have a couple of persnickety things here.

Also, for some reason your formatting is different than mine. I'm using clang-format version 10. I like mine 😁

src/sankoff_handler.cpp
```diff
@@ -104,14 +99,13 @@ void SankoffHandler::RunSankoff(Node::NodePtr topology) {
...
-            ParentPartial(psv_handler_.GetPV(PSVType::Q, node->Id())
-                              .block<state_count_, 1>(0, pattern_idx));
+            ParentPartial(psv_handler_.GetPV(PSVType::Q, node->Id()).col(pattern_idx));
         for (size_t child = 0; child < 2; child++) {
           size_t sister_idx = child == 0 ? 1 : 0;
           auto partials_from_sister = ParentPartial(
               TotalPPartial(node->Children()[sister_idx]->Id(), pattern_idx));
           psv_handler_.GetPV(PSVType::Q, node->Children()[child]->Id())
-              .block<state_count_, 1>(0, pattern_idx) =
+	      .col(pattern_idx) =
```
matsen, line 108: This looks like a tab but we're using spaces here.

src/sankoff_handler.hpp
```diff
@@ -13,9 +13,9 @@
 #include "driver.hpp"
 #include "pv_handler.hpp"
 
-// All partial vectors for all nodes at one site
+// Partial vector for one node across all sites
 using SankoffPartial = NucleotidePLV;
-// Each SankoffPartial represents a calculations for one site
+// Each SankoffPartial represents a calculations for one node
```
matsen, line 18:
```suggestion
// Each SankoffPartial represents calculations for one node
```

src/sankoff_handler.hpp
matsen: Did you want to do this, @marybarker ?

src/sankoff_handler.hpp
```diff
@@ -129,7 +133,7 @@ TEST_CASE("SankoffHandler: Tests on single site sequence.") {
       1., 2.5, 0., 2.5,       //
       2.5, 1., 2.5, 0.;       //
 
-  SankoffHandler sh = SankoffHandler(costs, site_pattern);
+  SankoffHandler sh = SankoffHandler(costs, site_pattern, "_ignore_mmaped_psv.data");
```
matsen, line 136: Was this supposed to be `_ignore/mmapped_psv.data`?

matsen (commented):
Looks nicer, Mary! It might be nice if you could respond to each of my comments, just saying "Fixed" or whatever and then I can resolve them as part of the PR approval process.

Do you want to do a less symmetric sankoff test?

src/sankoff_handler.cpp
marybarker: resolved using `Eigen::DenseBase::col`

src/sankoff_handler.cpp
marybarker: resolved

src/sankoff_handler.cpp
marybarker: removed

src/sankoff_handler.hpp
marybarker: resolved

src/sankoff_handler.hpp
marybarker: Ah, yes! That is on my list and I forgot. I'll check with Will D

src/sankoff_handler.hpp
marybarker: resolved

src/sankoff_handler.hpp
marybarker: resolved

src/sankoff_handler.cpp
marybarker: resolved

src/sankoff_handler.hpp
marybarker: there is now a test case with an asymmetric cost matrix.

matsen (approved):

PR #463, opened 2022-08-23
TPEngine: Computing Likelihoods with Proposed NNIs and NNIEngine Integration.

### LINKED ISSUE #461: TPEngine: Update after Add NNI

davidrich27:
## Description

- TPEngine
  - Adds function for computing likelihood for proposed NNI adjacent to DAG.
- NNIEngine
  - Adds un-owned reference to TPEngine and optionally adds TP-scoring schema and TPEngine maintenance routines.
  - Option added for NNIEngine to either or both TPEngine and GPEngine.
- GPInstance 
  - Adds capability to build and access TPEngine.

Closes #461

## Tests

TEST_CASE("Top-Pruning: Likelihoods with NNIs"): 
// Builds a TPEngine for two DAGs: a two-tree DAG, and a DAG formed from the original
// DAG plus all its adjacent NNIs. Both DAGs PVs are populated and their edge TP
// likelihoods are computed.  Then the first DAG's adjacent proposed NNI likelihoods are
// computed using only its own PVs.  Then compares the results of proposed NNIs from the
// first DAG with the known likelihoods from the second DAG.

## Checklist:


src/gp_instance.cpp
```diff
@@ -631,8 +633,29 @@ void GPInstance::SubsplitDAGToDot(const std::string &out_path,
...
+void GPInstance::MakeTPEngine(const std::string mmap_file_path,
+                              const bool using_likelihoods,
+                              const bool using_parsimony) {
+  auto site_pattern = MakeSitePattern();
+  tp_engine_ =
+      std::make_unique<TPEngine>(dag_, site_pattern, mmap_file_path, true, true);
+}
+
+TPEngine &GPInstance::GetTPEngine() {
+  Assert(tp_engine_, "GPInstance::GetTPEngine() when tp_engine has not been made.");
```
chrisjenningsshaffer, line 645:
```suggestion
  Assert(tp_engine_, "TpEngine not available. Call MakeTPEngine when tp_engine has not been made.");
```

src/gp_doctest.cpp
```diff
@@ -1989,3 +1987,100 @@ TEST_CASE("Top-Pruning: Likelihoods") {
       fasta_path_six, newick_path_six_simple, false, false);
   CHECK_MESSAGE(test_3, "Six Taxa Multi Tree failed.");
 }
+
+// Runs an instance of TPEngine for two DAGs: DAG_1, a simple DAG, and DAG_2, a DAG
+// formed from DAG_1 plus all of its adjacent NNIs. Both DAGs PVs are populated and
+// their edge TP likelihoods are computed.  Then DAG_1's adjacent proposed NNI
+// likelihoods are computed using only its own PVs. Then we compare the results of
+// the proposed NNIs from DAG_1 with the known likelihoods of DAG_2.
```
chrisjenningsshaffer, line 1995: Can you add a sentence saying what the comparison should show? The setup sounds reasonable, but I don't quite follow the last part.

chrisjenningsshaffer (changes requested):
Looks good. There are a couple minor things I marked in the comments.

src/nni_engine.hpp
```diff
@@ -252,13 +287,27 @@ class NNIEngine {
...
+  // Get Reference of GPEngine.
+  TPEngine &GetTPEngine() { return *tp_engine_; }
+
  private:
   // Un-owned reference DAG.
   GPDAG &dag_;
   // For adding temporary NNIs to DAG.
   std::unique_ptr<GraftDAG> graft_dag_;
+  // Whether NNIEngine is using or maintaining GPEngine.
+  bool using_gp_engine_ = false;
```
ognian-, line 304: Would it make sense to use std::optional<GPEngine*> instead of the bool + GPEngine* combination?

src/gp_doctest.cpp
davidrich27: Here's an update.  Is this clearer?

```
// Runs an instance of TPEngine for two DAGs: DAG_1, a simple DAG, and DAG_2, a DAG
// formed from DAG_1 plus all of its adjacent NNIs. Both DAGs PVs are populated and
// their edge TP likelihoods are computed.  Then DAG_1's adjacent proposed NNI
// likelihoods are computed using only PVs from the pre-NNI already contained in DAG_1.
// Finally, we compare the results of the proposed NNIs from DAG_1 with the known
// likelihoods of the actual NNIs already contained in DAG_2. This verifies we generate
// the same result from adding NNIs to the DAG and updating as we do from using the
// pre-NNI computation.
```

src/nni_engine.hpp
davidrich27: I was thinking I would give the option of setting a GPEngine and TPEngine, but only performing maintenance on them if they were being "used".  But I think that's probably an unnecessary complication that will rarely if ever have a use case.  Do you think optional with is better, or just use nullptr?

ognian-: I guess the simplest (nullptr) should work the best.

src/gp_doctest.cpp
chrisjenningsshaffer: Perfect.

chrisjenningsshaffer (approved):

src/nni_engine.hpp
davidrich27: Okay, changed it.

src/gp_doctest.cpp
```diff
@@ -1989,3 +1987,128 @@ TEST_CASE("Top-Pruning: Likelihoods") {
...
+    }
+    // Compute likelihoods with TPEngine.
+    for (const auto& [edge_id, tree_id] : tree_id_map) {
+      std::ignore = tree_id;
+      auto likelihood = tpengine.GetTopTreeLikelihoodWithEdge(edge_id);
+      tp_likelihood_map[edge_id] = likelihood;
+    }
+    return tp_likelihood_map;
+  };
+  // Compare likelihoods from TPEngine. DAG_1 must be a subset of DAG_2.
```
matsen, line 2045: Let's have a comment explaining exactly what's being tested here-- it's that each likelihood in the one DAG matches at least one other likelihood in the second dag, right?

src/gp_doctest.cpp
```diff
@@ -1989,3 +1987,128 @@ TEST_CASE("Top-Pruning: Likelihoods") {
       fasta_path_six, newick_path_six_simple, false, false);
   CHECK_MESSAGE(test_3, "Six Taxa Multi Tree failed.");
 }
+
+// Runs an instance of TPEngine for two DAGs: DAG_1, a simple DAG, and DAG_2, a DAG
```
matsen, line 1991: This is a great overall summary but it would be 🤩 if it was easier to connect it to the actual test below. Could you use the variable names here or change the variable names below to match? For example, I'm kind of lost what `new_likelihoods_without_nni` is in this description without reading everything in close detail.

matsen (commented):
This isn't something that has to get fixed immediately, but note that the `GPInstance` has sort of outgrown its name, because we can use it with top pruning as well. More important than the name, let's make sure the "identity" of this object still makes sense.

src/nni_engine.hpp
```diff
@@ -57,47 +63,79 @@ class NNIEngine {
...
+  void FilterPostUpdate();
+  // Apply the filtering method to determine whether each Adjacent NNI will be added to
+  // Accepted NNI or Rejected NNI.
+  void FilterProcessAdjacentNNIs();
+
+  // ** Filtering Schemes
+
+  // Set custom filtering scheme functions.
+  void SetFilteringScheme(
+      std::optional<StaticFilterInitFunction> filter_init_fn = std::nullopt,
```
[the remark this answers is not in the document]
matsen: But... thinking more broadly doesn't it make more sense to have separate setters for each of these?

matsen (commented):
Nice work! I look forward to learning more as the comments evolve.

src/gp_doctest.cpp
davidrich27: How's this?
```
  // Compare likelihoods from TPEngine. Every edge in DAG_1 must yield a tree likelihood
  // shared by at least one edge in DAG_2.
  ```

File: src/gp_doctest.cpp (hunk 5 above)
Comment (davidrich27) [reply src/gp_doctest.cpp:1991]: Would you prefer I used variables subscripted with 1 or 2 as opposed to with/without nni?  Instead of `new_likelihoods` maybe `proposed_likelihoods`?

Comment (matsen) [reply src/gp_doctest.cpp:1991]: Did our zoom conversation cover this?

File: src/gp_doctest.cpp (hunk 4 above)
Comment (davidrich27) [reply src/gp_doctest.cpp]: Updated.

File: src/gp_doctest.cpp (hunk 5 above)
Comment (davidrich27) [reply src/gp_doctest.cpp:1991]: Yes, it did.  Variable names updated.

File: src/nni_engine.cpp [hunk 7]
```diff
@@ -4,19 +4,18 @@
...
-#include "subsplit_dag.hpp"
-#include "nni_operation.hpp"
-#include "graft_dag.hpp"
-#include "sugar.hpp"
-
 using PLVType = PLVHandler::PLVType;
 
-NNIEngine::NNIEngine(GPDAG &dag, GPEngine &gp_engine)
-    : dag_(dag), graft_dag_(std::make_unique<GraftDAG>(dag)), gp_engine_(gp_engine) {}
+NNIEngine::NNIEngine(GPDAG &dag, std::optional<GPEngine *> gp_engine,
```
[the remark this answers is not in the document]
davidrich27: Yes, changed.

src/nni_engine.cpp
```diff
@@ -62,17 +61,128 @@ void NNIEngine::RunPostLoop() {
...
 // ** Filter Functions
 
-double NNIEngine::NoEvaluate(NNIEngine &this_nni_engine, GPEngine &this_gp_engine,
-                             GraftDAG &this_graft_dag, const NNIOperation &nni) {
-  return 0.0;
+void NNIEngine::SetNoEvaluate() {
+  // NNI Evaluation function
+  filter_process_fn_ = [](NNIEngine &this_nni_engine, GPEngine &this_gp_engine,
+                          TPEngine &this_tp_engine, GraftDAG &this_graft_dag,
+                          const NNIOperation &nni) -> double { return 0.0; };
```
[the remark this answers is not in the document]
davidrich27: Nope, that was a mistake.  It is a different function.

src/nni_engine.cpp
```diff
@@ -62,17 +61,128 @@ void NNIEngine::RunPostLoop() {
...
 }
 
-bool NNIEngine::NoFilter(NNIEngine &this_nni_engine, GPEngine &this_gp_engine,
-                         GraftDAG &this_graft_dag, const NNIOperation &nni) {
-  return true;
+void NNIEngine::SetNoFilter(const bool set_nni_to_pass) {
+  // NNI Processing function
```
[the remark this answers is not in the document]
davidrich27: Removed internal comments and added further explanation of each filter function's purpose in the header.

src/nni_engine.hpp
```diff
@@ -57,47 +63,79 @@ class NNIEngine {
...
   void RunMainLoop();
+  // Step that runs at the end of each loop, preps for next loop.
   void RunPostLoop();
 
   // ** Filter Functions
 
   // Initialization step for filter before beginning
   using StaticFilterInitFunction =
-      std::function<void(NNIEngine &, GPEngine &, GraftDAG &)>;
+      std::function<void(NNIEngine &, GPEngine &, TPEngine &, GraftDAG &)>;
```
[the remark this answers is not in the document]
davidrich27: That is a good point.  I think we should add declare a some common GP/TP/XYEngine API interface, then NNI Engine can hold one or more of those engines and rather than NNI Engine storing any maintenance logic, can just make direct calls to this interface.

src/nni_engine.hpp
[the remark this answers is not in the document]
davidrich27: Yeah, splitting the function up seems reasonable.  Done.

src/tp_engine.cpp
```diff
@@ -71,9 +71,112 @@ void TPEngine::InitializeLikelihood() {
...
+  CopyBranchLengthFromCommonAdjacentNodes(pre_parent_id, post_parent_id,
+                                          Direction::Rootward, SubsplitClade::Right);
+  CopyBranchLengthFromCommonAdjacentNodes(
+      pre_parent_id, post_child_id, Direction::Leafward,
+      Bitset::Opposite(pre_edge.GetSubsplitClade()));
+  // Copy over left and right edges.
+  NodeId post_leftchild_id;
+  NodeId post_rightchild_id;
+  if (pre_nni.GetSisterClade() == post_nni.GetLeftChildClade()) {
+    // If post_nni swapped pre_nni sister swapped with pre_nni left child.
```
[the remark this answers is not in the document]
davidrich27: The post-NNI is the result of the pre-NNI performing a swap between its parent sister clade and either it left child or its right child.  This function is checking which it is by finding which cladeside of the post-NNI child matches the pre-NNI sister.

src/nni_engine.hpp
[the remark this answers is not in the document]
matsen: Great. Keep that thought going but let's stay the course for now.

src/tp_engine.cpp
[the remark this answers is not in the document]
matsen: So it seems like a relevant comment would be

```suggestion
    // If post_nni swapped pre_nni sister with pre_nni left child.
```

Is that right? If so, can you correct this one and the ones below? Thx

src/gp_doctest.cpp
matsen: Marvelous, thanks! Much clearer now.

src/tp_engine.cpp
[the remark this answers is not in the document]
davidrich27: Yes, that is correct.  Fixed.

PR #477, opened 2023-06-22
Ordering Taxon IDs

### LINKED ISSUE #476: Rooted trees ignore original taxa labels
Some operations on rooted trees use certain internal representations for calculations that should instead use taxa labels. 

This can cause unwanted behavior like reading the newick string "(2,(1,3))" and writing out the newick string "(0,(1,2))",
not recognizing the newick strings "((1,2),3)" and "(3,(1,2))" represent the same topology, and not recognizing
the newick strings "(1,(2,3))" and "(2,(1,3))" represent different topologies.

A self-contained python script demonstrating these issues is at:
`/fh/fast/matsen_e/shared/vip/temp/taxon_order_issues/taxa_labels_and_order.py`

I have not checked if the same issues come up for unrooted tree instances.

davidrich27:
## Description

- When building trees from Newick files previously, taxon ids were set according to the order of appearance of each taxon in the first Newick string.  This creates issues with two identical rooted trees appearing different, because branches can rotate due to bitset ordering.
- Updated so now taxon ids are assigned alphabetically.  So now different Newick strings with the same topology should be correctly identified.
- Added a flag option when reading in files whether to sort taxon ids or not.

Closes #476 

## Tests

- GPInstance: Taxon Sorted Tree Collection
  - Builds the same tree from two different Newick strings: "(1,(2,3))'" and "((2,3),1)".  
    - First, builds trees without sorting flag set and asserts trees are not equal.
    - Second, builds trees with sorting flag set and asserts trees are equal.

## Checklist:

* [ ] Issue branch has been squashed and rebased on main branch


src/driver.cpp
chrisjenningsshaffer, line 1: Looks good.

I like Erick's suggestion of making this version optional, with the user setting a flag to indicate they want taxon loaded this way. Other than that addition, this looks good to go.

chrisjenningsshaffer (commented):
Looks good.

I like Erick's suggestion of making this version optional, with the user setting a flag to indicate they want taxon loaded this way. Other than that addition, this looks good to go.

matsen:
Yep, I'd like to review after adding the taxon sorting flag. Thx!

src/driver.cpp
```diff
@@ -201,4 +208,20 @@ TagStringMap Driver::TagTaxonMap() {
...
 
+void Driver::SetTaxa(const std::map<std::string, uint32_t> taxa) {
+  taxa_ = taxa;
+  taxa_complete_ = true;
+}
+
+void Driver::SortTaxa() {
+  std::map<std::string, uint32_t> taxa;
+  uint32_t new_id = 0;
+  for (const auto &[name, old_id] : taxa_) {
```
matsen, line 219: Is this guaranteed to iterate in alphabetical order?

davidrich27: Yes, std::map is guaranteed to iterate over keys in alphabetical order.

https://stackoverflow.com/questions/7648756/is-the-order-of-iterating-through-stdmap-known-and-guaranteed-by-the-standard

matsen (approved):

PR #481, opened 2023-09-06
Adds mapping layer for PVHandler

davidrich27:
## Description

Another proposed speedup for NNI search.   Currently, whenever the DAG is modified, the PVs are reindexed by shuffling the data vector to be ordered with respect to their node_ids or edge_ids.  This reordering requires copying nearly all of the PVs in the DAG, creating a sizeable time cost.  This change creates a map between PV id, and PV location, for a lightweight way to reindex the PVs.  The full reordering option is left in, because as long as it is an amortized cost, it can be helpful in preventing the PVs from becoming overly fragmented for access patterns. 

- Adds method to PLVHandler for reindexing via remapping.
- Adds an option to PLVHandler to reindex via remapping or copying.
- Adds updates to the sDAG clade maps so that clade maps are also used for finding grafted node neighbors (before this was only for host nodes).

## Tests

- Updates to TEST_CASE("NNIEngine: Resize and Reindex GPEngine after AddNodePair") test this updated functionality.
- TEST_CASE("TPEngine: Resize and Reindex PV Handler"):  Iteratively adds NNIs to two engines, one engine using the old update via copying function and one using the new update via remapping function.  Tests that each method have the same PV values at each step as NNIs are added to the DAG. 

## Checklist:


matsen (changes requested):
Thanks for this, Dave! It's so great that you found this problem and then developed a complete fix for it. 

If you want me to look this over, could you make one more pass to check for cruft? There are some extra files and other bits. It looks like it could use some rebasing as well.

There's no rush for this...

matsen (approved):
Looks great, Dave! Thanks for the thorough tests.

